PR #3, opened 2020-02-18
Use no-std-net definitions of IpAddr and SocketAddr

### LINKED ISSUE #1: Use existing crates
This looks great for getting us started!

Maybe we should consider using the existing https://github.com/dunmatt/no-std-net for IPAddress and SocketAddress, as that is essentially the `std::net` without the `std` part.

Perhaps it would even make sense eventually to merge the two..

MathiasKoch:
Fixes #1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,4 +10,5 @@ license = "MIT or Apache-2.0"
 
 [dependencies]
 nb = "0.1"
-no-std-net = "0.2.3"
+# no-std-net = "0.2.3"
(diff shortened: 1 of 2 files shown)


thejpster:
I'm afraid I'm not convinced mutable references are the way to go.

MathiasKoch:
I agree, will revert that one commit first thing tomorrow, and update this PR 👍

thejpster (approved):

thejpster:
Thanks!

PR #9, opened 2020-05-07
First edition of proposal for dns trait

### LINKED ISSUE #8: Support trait for DNS
Would it make sense to add a support trait for DNS lookups?

Somthing along the lines of
```rust
pub trait Dns {
    /// The type returned when we have an error
    type Error: core::fmt::Debug;

    fn resolve(&self, hostname: &str) -> Result<IpAddr, Self::Error>;
}
```

MathiasKoch:
This PR provides a simple DNS trait, similar to that provided by both `lwip` and the `Berkeley sockets API`, consisting of a simple `gethostbyname` function that resolves a hostname to the first ip address given by `A` or `AAAA` DNS record, depending on the requested `addr_type`, and a `gethostbyaddr` to resolve the hostname given an ip address. The later function has the `addr_type` encoded in the given `IpAddr` input.

This is a proposal for a first edition. It will not cover the full DNS, but it is equivalent of both `lwip` and the `Berkeley sockets API`, as well as the functionality given by `std::net::ToSocketAddrs`

Fixes #8

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,5 +1,5 @@
+use heapless::{consts, String};
 use no_std_net::IpAddr;
-use heapless::{String, consts};
 
 /// This is the host address type to be returned by `gethostbyname`.
 ///
@@ -7,10 +7,12 @@ use heapless::{String, consts};
 /// will look for `AAAA` records
 #[derive(Clone, Debug, PartialEq)]
 pub enum AddrType {
-    /// Result is `A` record
-    IPv4,
-    /// Result is `AAAA` record
-    IPv6
+	/// Result is `A` record
+	IPv4,
+	/// Result is `AAAA` record
+	IPv6,
+	/// Result is either a `A` record, or a `AAAA` record
+	Either,
 }
 
 /// This trait is an extension trait for [`TcpStack`] and [`UdpStack`] for dns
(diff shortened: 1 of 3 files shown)


src/dns.rs
```diff
@@ -0,0 +1,41 @@
+use no_std_net::IpAddr;
+use heapless::{String, consts};
+
+/// This is the host address type to be returned by `gethostbyname`.
+///
+/// An IPv4 address type always looks for `A` records, while IPv6 address type
+/// will look for `AAAA` records
+#[derive(Clone, Debug, PartialEq)]
+pub enum AddrType {
```
thejpster, line 9: Is there anything in `no_std_net` we can use here. Feels like we're re-inventing the wheel?

thejpster: Also, should we support an 'Either' option, in case you support both and don't mind either? I think you can ask a DNS sever for both AAAA and A records at the same time, so this would be more efficient than asking for one then the other.

thejpster (changes requested):
Just a comment on the AddrType type.

src/dns.rs
MathiasKoch: I had sort of the same feeling, but they don't define anything like this. All they have is the `IpAddr` and `SocketAddr`, and the variants in both hold `Ipv4Addr/Ipv6Addr` and `SocketAddrV4/SocketAddrV6`.

> Also, should we support an 'Either' option, in case you support both and don't mind either? I think you can ask a DNS sever for both AAAA and A records at the same time, so this would be more efficient than asking for one then the other.

I think this makes sense. An alternative could be to change `addr_type` in `gethostbyname` to `Option<AddrType>`? Where `None` actually means either. Not sure which is better?

MathiasKoch: @thejpster Did you have any opinions on adding an `Either` vs changing it to an `Option`? See above comment.

thejpster: Sorry, I'm not on Github much any more. I think I prefer `Either|IPv6|IPv4` over `Some(IPv4|IPv6)|None`, as `None` sounds you like don't want either, as opposed to both.

thejpster (approved):

thejpster:
LGTM!

PR #13, opened 2020-08-26
Release 0.1.0

### LINKED ISSUE #11: Publish crate to crates.io
Currently, I know both I and @MathiasKoch are looking to publish crates that utilize the `nal` to crates.io, but that's currently not possible because the `embedded-nal` has not yet been published.

Is it possible for us to generate an initial release for this crate, and/or are there things we should consider before publishing?

eldruin:
Among other small things I have updated `no-std-net` to version `0.4`. Is that a problem?

I have also added some basic information about how to propose new traits and so on borrowed from `embedded-hal` but without defining the demonstration requirements, since this is more of a playground.

I have marked 1.36.0 as the MSRV because that is the lowest one compatible. Would you prefer a higher version?

cc. @ryan-summers, @MathiasKoch

Fixes #11

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,5 +10,5 @@ license = "MIT or Apache-2.0"
 
 [dependencies]
 nb = "1"
-no-std-net = "0.3.0"
+no-std-net = "0.4"
 heapless = "^0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,11 @@
 [package]
 name = "embedded-nal"
 version = "0.1.0"
-authors = ["Jonathan 'theJPster' Pallant <github@thejpster.org.uk>"]
+authors = [
+    "Jonathan 'theJPster' Pallant <github@thejpster.org.uk>",
+    "Mathias Koch <mk@blackbird.online>",
+    "Diego Barrios Romero <eldruin@gmail.com>",
+]
 edition = "2018"
 description = "A Network Abstraction Layer (NAL) for Embedded Systems"
 license = "MIT or Apache-2.0"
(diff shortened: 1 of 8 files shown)


MathiasKoch:
LGTM! Thanks!

Cargo.toml
```diff
@@ -1,14 +1,22 @@
...
+homepage = "https://github.com/rust-embedded-community/embedded-nal"
+documentation = "https://docs.rs/embedded-nal"
+readme = "README.md"
+keywords = ["network"]
+categories = ["embedded", "hardware-support", "no-std", "network-programming"]
 
 [dependencies]
-nb = "0.1"
-no-std-net = "0.3.0"
+nb = "1"
```
ryan-summers, line 20: One thing I noticed during development is that the traits require `nb::Result`, but `nb::Result` (or `nb`) is not re-exported by the crate, which can make a weird dependency situation. Can we add a `pub use nb;` to `lib.rs`?

We can also push this off into an issue and leave as-is, it's just a general comment that I noticed. I know `nb` doesn't change much so it's not a huge inconsistency as-is.

ryan-summers (changes requested):

Cargo.toml
eldruin: Sure thing. I added it now. thanks!

eldruin:
How about some tests or doctests? Can you think of something useful?

ryan-summers:
The embedded-hal similarly does not have any tests / doctests that I can see and it might be a bit burdensome to add tests, since we'd need to define a whole network stack. Perhaps we can provide a sample implementation using `std::net` to implement the traits? I would recommend we push this into an issue.

Edit: I actually made a reference impl for std::net here: https://github.com/quartiq/minimq/blob/master/tests/integration_test.rs

ryan-summers (approved):

MathiasKoch (approved):

eldruin:
Alright. Let's get this released then!

Issue #4, opened 2020-02-18
self mutable reference


MathiasKoch:
I am a bit unsure on why this shouldn't have a mutable reference to self?

The way i see it, you would usually implement this on some sort of network client, owning access to a peripheral (unix fd, SPI, I2C, Serial etc.), but most (all?) of these require a mutable reference to be able to send data. This could be fixed by wrapping the peripheral in a `RefCell`, but even then there is a good chance that the client object would hold stuff like internal state etc that needs to be mutated, meaning that in my particular case, i would end up with a client object of 5 fields, all wrapped in `RefCell`.

Is this intended behavior, or am i implementing something wrong in general?

thejpster:
I had intended implemetors of the crates to handle internal mutability. While in the BSD socket library, I can call `open()` or `write()` etc, without having to hold any kind of pointer or reference - it is handled internally by the OS - here, having to hold a non-mutable reference to a particular IP stack seemed like a good compromise.

Requiring a `&mut ref` would make operating, say, an HTTP client and an SMTP client at the same time almost impossible.

MathiasKoch:
I think that makes sense. I will go with the internal mutability :+1:

kellerkindt:
Cross linking to #39

PR #21, opened 2020-11-11
Added UDP Server Trait

### LINKED ISSUE #20: [Question] No way to set UDP port?
I see that SocketAddr supports an IP address and a remote port number.  However, I see no way to specify a local port number.  How would a program specify the port to listen on/send from without that ability?

jonahd-g:
/resolves #20

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,11 +65,19 @@ pub trait UdpStack {
 	/// The type returned when we have an error
 	type Error: core::fmt::Debug;
 
-	/// Open a local port on a new UDP socket to the given address and port.
+	/// Open a UDP socket with a dynamically selected port.
+	///
+	/// Selects a port number automatically and opens a UDP socket to the given
+	/// address. UDP is connectionless, so unlike `TcpStack` no `connect()` is
+	/// required.
+	fn open(&self, remote: SocketAddr, mode: Mode) -> Result<Self::UdpSocket, Self::Error>;
+
+	/// Open a new UDP socket with a specified port
+	///
+	/// Opens a new socket with the specified port number to the given address.
 	/// UDP is connectionless, so unlike `TcpStack` no `connect()` is required.
-	fn open(
+	fn bind(
 		&self,
-		local_port: u16,
 		remote: SocketAddr,
 	) -> Result<Self::UdpSocket, Self::Error>;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,21 +65,16 @@ pub trait UdpStack {
 	/// The type returned when we have an error
 	type Error: core::fmt::Debug;
 
-	/// Create a new socket
-	///
-	/// Allocates a socket for use as a client or a server.
-	fn socket(&self) -> Result<Self::UdpSocket, Self::Error>;
-
 	/// Open a UDP socket with a dynamically selected port.
 	///
 	/// Selects a port number automatically and initializes for read/writing.
-	fn connect(&self, socket: &mut Self::UdpSocket, remote: SocketAddr) -> Result<(), Self::Error>;
+	fn connect(&self, remote: SocketAddr) -> Result<Self::UdpSocket, Self::Error>;
 
 	/// Send a datagram to the remote host.
 	///
 	/// The remote host used is either the one specified in `UdpStack::connect`
 	/// or the last one used in `UdpServerStack::write_to`.
-	fn write(&self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error>;
+	fn send(&self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error>;
 
 	/// Read a datagram the remote host has sent to us.
 	///
@@ -88,7 +83,7 @@ pub trait UdpStack {
 	/// If a packet has not been received when called, then
 	/// [`nb::Error::WouldBlock`](https://docs.rs/nb/1.0.0/nb/enum.Error.html#variant.WouldBlock)
 	/// should be returned.
-	fn read(
+	fn receive(
 		&self,
 		socket: &mut Self::UdpSocket,
 		buffer: &mut [u8],
@@ -104,10 +99,10 @@ pub trait UdpServerStack: UdpStack {
 	/// Open a new UDP socket with a specified port
 	///
 	/// Opens a new socket with the specified port number to the given address.
-	fn bind(&self, socket: &mut Self::UdpSocket, local_port: u16) -> Result<(), Self::Error>;
+	fn bind(&self, local_port: u16) -> Result<Self::UdpSocket, Self::Error>;
 
 	/// Send a packet to a remote host/port.
-	fn write_to(
+	fn send_to(
 		&self,
 		socket: &mut Self::UdpSocket,
 		remote: SocketAddr,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,23 +65,20 @@ pub trait UdpStack {
 	/// The type returned when we have an error
 	type Error: core::fmt::Debug;
 
+	/// Create a new socket
+	///
+	/// Allocates a socket for use as a client or a server.
+	fn socket(&self) -> Result<Self::UdpSocket, Self::Error>;
+
 	/// Open a UDP socket with a dynamically selected port.
 	///
-	/// Selects a port number automatically and opens a UDP socket to the given
-	/// address. UDP is connectionless, so unlike `TcpStack` no `connect()` is
-	/// required.
-	fn open(&self, remote: SocketAddr, mode: Mode) -> Result<Self::UdpSocket, Self::Error>;
-
-	/// Open a new UDP socket with a specified port
-	///
-	/// Opens a new socket with the specified port number to the given address.
-	/// UDP is connectionless, so unlike `TcpStack` no `connect()` is required.
-	fn bind(
-		&self,
-		remote: SocketAddr,
-	) -> Result<Self::UdpSocket, Self::Error>;
+	/// Selects a port number automatically and initializes for read/writing.
+	fn connect(&self, socket: &mut Self::UdpSocket, remote: SocketAddr) -> Result<(), Self::Error>;
 
 	/// Send a datagram to the remote host.
+	///
+	/// The remote host used is either the one specified in `UdpStack::connect`
+	/// or the last one used in `UdpServerStack::write_to`.
 	fn write(&self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error>;
 
 	/// Read a datagram the remote host has sent to us.
@@ -99,6 +96,23 @@ pub trait UdpStack {
 
 	/// Close an existing UDP socket.
 	fn close(&self, socket: Self::UdpSocket) -> Result<(), Self::Error>;
+}
+
+/// This trait is implemented by UDP/IP stacks.  It provides the ability to
+/// listen for packets on a specified port and send replies.
+pub trait UdpServerStack: UdpStack {
+	/// Open a new UDP socket with a specified port
+	///
+	/// Opens a new socket with the specified port number to the given address.
+	fn bind(&self, socket: &mut Self::UdpSocket, local_port: u16) -> Result<(), Self::Error>;
+
+	/// Send a packet to a remote host/port.
+	fn write_to(
+		&self,
+		socket: &mut Self::UdpSocket,
+		remote: SocketAddr,
+		buffer: &[u8],
+	) -> nb::Result<(), Self::Error>;
 }
 
 // End Of File
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,7 +59,7 @@ pub trait TcpStack {
 /// module. You could have another implemenation which knows how to driver the
 /// Rust Standard Library's `std::net` module. Given this trait, you can how
 /// write a portable CoAP client which can work with either implementation.
-pub trait UdpStack {
+pub trait UdpClient {
 	/// The type returned when we create a new UDP socket
 	type UdpSocket;
 	/// The type returned when we have an error
@@ -95,7 +95,7 @@ pub trait UdpStack {
 
 /// This trait is implemented by UDP/IP stacks.  It provides the ability to
 /// listen for packets on a specified port and send replies.
-pub trait UdpServerStack: UdpStack {
+pub trait UdpServer: UdpClient {
 	/// Open a new UDP socket with a specified port
 	///
 	/// Opens a new socket with the specified port number to the given address.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,9 +65,13 @@ pub trait UdpStack {
 	/// The type returned when we have an error
 	type Error: core::fmt::Debug;
 
-	/// Open a new UDP socket to the given address and port. UDP is connectionless,
-	/// so unlike `TcpStack` no `connect()` is required.
-	fn open(&self, remote: SocketAddr) -> Result<Self::UdpSocket, Self::Error>;
+	/// Open a local port on a new UDP socket to the given address and port.
+	/// UDP is connectionless, so unlike `TcpStack` no `connect()` is required.
+	fn open(
+		&self,
+		local_port: u16,
+		remote: SocketAddr,
+	) -> Result<Self::UdpSocket, Self::Error>;
 
 	/// Send a datagram to the remote host.
 	fn write(&self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error>;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
-- Ability to specify a local UDP port
+- Added a new `UdpServer` trait with server-specific methods
 ### Changed
 - Changed the `UdpStack::receive` method to return the packet sender address, along with the packet length
+- Changed the name of `UdpStack` to `UdpClient`
 ### Removed
 - Removed `Mode` enum, implementations should instead use `nb::WouldBlock`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Ability to specify a local UDP port
 ### Changed
 - Changed the `UdpStack::receive` method to return the packet sender address, along with the packet length
 ### Removed


Sympatron:
How do you open without selecting a local port yourself then?
Maybe it would be better to make `local_port` an `Option<u16>`?

jonahd-g:
Do you mean that the interface should support letting the driver automatically select a port automatically?

Sympatron:
That's how the API was designed previously and that's also how the POSIX socket API works. I think it is useful if you don't care about your local port number.

ryan-summers (commented):
- [ ] Please add a changelog entry for this as well :)

jonahd-g:
@Sympatron I implemented the optional port as a different overload, lmk if you want to discuss that choice.

@ryan-summers Are you sure we want another release tag for this one change?  There should be some more changes coming down the pipe in my other PRs.

ryan-summers:
> @ryan-summers Are you sure we want another release tag for this one change? There should be some more changes coming down the pipe in my other PRs.

This should be added as a bullet into the `Unreleased` section of the changelog. That way, when we _do_ generate a release, it's easy for us to see what has changed since the last release.

MathiasKoch:
@jonahd-g You can't do overloads like that in Rust.. You need to either make it `Option<>`, or make a new function with a new name

jonahd-g:
> This should be added as a bullet into the Unreleased section of the changelog. That way, when we do generate a release, it's easy for us to see what has changed since the last release.

Lol I realized that was probably the case lying in bed last night.  Will do.

> You can't do overloads like that in Rust.. You need to either make it Option<>, or make a new function with a new name

Of course, thank you.  My Rust is a little... rusty :sunglasses:

jonahd-g:
:heavy_check_mark:

src/lib.rs
```diff
@@ -71,10 +71,24 @@ pub trait UdpStack {
...
+	/// Selects a port number automatically and opens a UDP socket to the given
+	/// address. UDP is connectionless, so unlike `TcpStack` no `connect()` is
+	/// required.
 	fn open(&self, remote: SocketAddr, mode: Mode) -> Result<Self::UdpSocket, Self::Error>;
 
+	/// Open a new UDP socket with a specified port
+	///
+	/// Opens a new socket with the specified port number to the given address.
+	/// UDP is connectionless, so unlike `TcpStack` no `connect()` is required.
+	fn open_port(
```
ryan-summers: If we're following specific networking semantics, I think we would actually want to create a bind function. It doesn't make sense to have to different open functions.

It's also worth questioning why this would be required in the embedded-nal. Keep in mind that the more functions we add to the trait, the harder it is for users to implement and the less likely people will be to use the `embedded-nal` interface. We should be careful about adding more functions than necessary and over-complicating the trait.

That being said, if we did want this functionality, I believe the proper approach would be to add a `bind` function to bind the socket to a specific port:

```rust
fn bind(&self, socket: Self::UdpSocket, address: SocketAddr) -> Result<Self::UdpSocket, Self::Error>;
```

ryan-summers (changes requested):

src/lib.rs
ryan-summers: @jonahd-g Would you please provide some sample use cases as to where `bind()`-like functionality would be needed? Is it possible that your use case can be done _outside_ of the NAL?

jonahd-g: @ryan-summers respectfully, the need to bind to a specific port is such an obviously fundamental requirement of a NAL to me that I don't really understand the question.  Like, if I want a UDP server, I need to be able to listen on a particular port, otherwise nobody knows where to find it.

I'm fine with naming it bind, that makes sense.  And even if you want to go back to the `Option` type arg I'd be okay with that too.

ryan-summers: I totally spaced that we can have embedded applications that are servers as well as clients. Bind totally makes sense here. Feel free to ignore my short-sightedness :)

That being said, I think from an API simplicity level, I'd prefer this being an `Option` in `open()`, but I'm noticing that my opinion seems to be shifting back and forth here.

Arguments for an external `bind()` function:
* It's logical and follows other network stacks

Arguments for an `Option<u16>`:
* It simplifies the trait so that it's easier to implement.

jonahd-g: > I totally spaced that we can have embedded applications that are servers as well as clients. Bind totally makes sense here. Feel free to ignore my short-sightedness :)

Haha sounds good.

> That being said, I think from an API simplicity level, I'd prefer this being an Option in open(), but I'm noticing that my opinion seems to be shifting back and forth here.

Yes I'm of two minds on this one as well.  The reason I wrote it as a separate method is so that the flow is known at compile time.  If the user only calls the `bind` method, the compiler can omit the `open` method entirely.  `Open` would be pretty much just `make up a port and call bind(port)`.  Although, perhaps the compiler can _already_ statically analyze the usage.  I don't know what kind of optimizations it makes, this could be a moot point.

MathiasKoch: I would personally lean towards following the Berkeley socket API https://en.wikipedia.org/wiki/Berkeley_sockets 

That way we follow some sort of standard, and one that is widely used in C & POSIX, making adoption easier as it looks and feels familiar to devs.

jonahd-g: @MathiasKoch What would be the list of methods if we did that?  My reading is that it should be:

- socket
- bind
- read
- write
- close

MathiasKoch: Basically yes. 
Where we currently have `open` roughly equivalent to `socket` + `connect` when talking UDP client

jonahd-g: Ahh, so `connect` would be the antithesis of `bind`.  I like that, I'll make that revision.

MathiasKoch: :+1:

jonahd-g:
@MathiasKoch I think we might need a `sendto` function too.  If a server receives a packet, often it's necessary to send a packet back to that sender.  We don't want a program to need to `release()` then `bind` all over again just to change the destination address.  Thoughts on this?

Sympatron:
I think `bind` shouldn't take a remote address at all. If you want to have a UDP server, you don't know the remote address yet.
Instead something like `sendto` should exist to use this socket to send to some remote address.

MathiasKoch:
I don't know if it could make sense to split both traits into a core and a server trait, where the server trait could be considered optional?

```rust
pub trait UdpCore {
	/// The type returned when we create a new UDP socket
	type UdpSocket;
	/// The type returned when we have an error
	type Error: core::fmt::Debug;

	fn open(&self, remote: SocketAddr, mode: Mode) -> Result<Self::UdpSocket, Self::Error>;

	fn write(&self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error>;

	fn read(
		&self,
		socket: &mut Self::UdpSocket,
		buffer: &mut [u8],
	) -> nb::Result<(usize, SocketAddr), Self::Error>;

	fn close(&self, socket: Self::UdpSocket) -> Result<(), Self::Error>;
}

pub trait UdpServer: UdpCore {
	fn bind(&self, socket: <Self as UdpCore>::UdpSocket, local_port: u16) -> Result<(), <Self as UdpCore>::Error>;

    .... More UDP server functions? eg. sendto ?
}
```

It might look like its very few functions in the UDP case, but its a few more in TCP (accept, listen, bind...)

Sympatron:
That may solve the "too many function to implement" problem stated above. I like it.

MathiasKoch:
> I think `bind` shouldn't take a remote address at all. If you want to have a UDP server, you don't know the remote address yet.
> Instead something like `sendto` should exist to use this socket to send to some remote address.

This is the Berkeley definition of `bind`:
```
bind() associates a socket with an address. When a socket is created with socket(), it is only given a protocol family, but not assigned an address. This association must be performed before the socket can accept connections from other hosts. The function has three arguments:

- sockfd, a descriptor representing the socket
- my_addr, a pointer to a sockaddr structure representing the address to bind to.
- addrlen, a field of type socklen_t specifying the size of the sockaddr structure.
Bind() returns 0 on success and -1 if an error occurs.
```

MathiasKoch:
> That may solve the "too many function to implement" problem stated above. I like it.

Exactly my though.. Especially considering that many will want the client side (or core), but few will require the server side.

MathiasKoch:
@Sympatron Think you're misunderstanding `bind`.. it doesn't take a remote port, but rather a local port to accept incoming datagrams on

Sympatron:
> @Sympatron Think your misunderstanding `bind`.. it doesn't take a remote port, but rather a local port to accept incoming datagrams on

I was not talking about my understanding, but about the [current state of the PR](https://github.com/rust-embedded-community/embedded-nal/pull/21/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R91).

MathiasKoch:
> > @Sympatron Think your misunderstanding `bind`.. it doesn't take a remote port, but rather a local port to accept incoming datagrams on
> 
> I was not talking about my understanding, but about the [current state of the PR](https://github.com/rust-embedded-community/embedded-nal/pull/21/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R91).

Ahh, okay.. Sorry about that. In that case i totally agree!

jonahd-g:
> I don't know if it could make sense to split both traits into a core and a server trait, where the server trait could be considered optional?

I disagree that it should be considered "optional".  If drivers don't implement it, they should be considered incomplete, IMO.

> `fn bind(&self, socket: <Self as UdpCore>::UdpSocket, local_port: u16) -> Result<(), <Self as UdpCore>::Error>;`

I didn't incorporate the `Self as UdpCore` into my PR. Can you explain why it should be?

Sympatron:
I think `embedded-nal` should follow Berkeley sockets where appropriate, because it is well known and proven. On the other hand I don't know why they split the initialization into 2 parts. As far as I can see you always have to call either `socket()` + `connect()` or `socket()` + `bind()`. `socket()` on it's own is quite useless and returns an object in a mostly unusable state. Therefore I would vote to combine them in `connect()` and `bind()`. Unlike `socket()` the meaning would be unambiguous and would return an object in a valid usable state.

I would also vote for `send_to()` instead of `write_to()` (and also `receive_from()` instead of `read()`) to follow the standard more closely, but I guess this comes down to personal preference.

jonahd-g:
I agree, I didn't want to presume to rename the read/write methods.

> (and also receive_from() instead of read())

Why receive_from instead of simply receive?  What benefit would that provide?

ryan-summers:
> > I don't know if it could make sense to split both traits into a core and a server trait, where the server trait could be considered optional?
> 
> I disagree that it should be considered "optional". If drivers don't implement it, they should be considered incomplete, IMO.

I have a few client-only stack implementations and tend to think it's actually the "standard use case" - for example, MQTT or DNS clients. Servers tend to usually reside on linux- or windows-hosts, since they're usually more complicated and connected to cloud-based back-ends etc. I was personally thinking of splitting it into `Client` and `Server` traits as well, and I like the idea.

If you want something that implements both a UDP server + client, you would just do:

```rust
pub fn my_server_function<T: UdpServer + UdpClient>(stack: T);
```

MathiasKoch:
> > > I don't know if it could make sense to split both traits into a core and a server trait, where the server trait could be considered optional?
> > 
> > 
> > I disagree that it should be considered "optional". If drivers don't implement it, they should be considered incomplete, IMO.
> 
> I have a few client-only stack implementations and tend to think it's actually the "standard use case" - for example, MQTT or DNS clients. Servers tend to usually reside on linux- or windows-hosts, since they're usually more complicated and connected to cloud-based back-ends etc. I was personally thinking of splitting it into `Client` and `Server` traits as well, and I like the idea.
> 
> If you want something that implements both a UDP server + client, you would just do:
> 
> ```rust
> pub fn my_server_function<T: UdpServer + UdpClient>(stack: T);
> ```

I have the same, lots of client only implementations, but the server part would be a whole other headache to implement and maintain for me, as i am not using it anywhere. That said i would love to open up for the possibility of someone contributing the server implementation afterwards.

WRT `Client` and `Server` traits, i just figured you can't really have the server part without implementing the client functions (i guess?), thus why i went with a `Core` and `Server` trait separation

jonahd-g:
> I have a few client-only stack implementations and tend to think it's actually the "standard use case" - for example, MQTT or DNS clients. Servers tend to usually reside on linux- or windows-hosts, since they're usually more complicated and connected to cloud-based back-ends etc. I was personally thinking of splitting it into Client and Server traits as well, and I like the idea.

I think we may be down to semantics at this point.  IMO it should be perfectly fine for an implementation to be incomplete, but it should be recognized as such to avoid users making any assumptions about what features it has.  But if I'm outnumbered and the server trait is "optional", I'm won't lose any sleep over it. :smiley: 

> WRT Client and Server traits, i just figured you can't really have the server part without implementing the client functions (i guess?), thus why i went with a Core and Server trait separation

Are you saying you want the PR to rename UdpStack to UdpCore?  Is the TCP trait expected to similarly give up usage of the word "stack" at some point?

MathiasKoch:
> Are you saying you want the PR to rename UdpStack to UdpCore? Is the TCP trait expected to similarly give up usage of the word "stack" at some point?

I don't think that is for me to decide, but I think we should iron out breaking changes like this asap, if we want to rename/restructure the traits themselves. In case we choose to change the naming, I would definitely say it should be reflected in the TCP version as well yes! And especially in this case, where the server split is at least as applicable in the TCP version.

ryan-summers:
> I think we may be down to semantics at this point. IMO it should be perfectly fine for an implementation to be incomplete, but it should be recognized as such to avoid users making any assumptions about what features it has. But if I'm outnumbered and the server trait is "optional", I'm won't lose any sleep over it. 😃

The idea here is that we would then make it clear to the user that something doesn't implement a `UdpServer` at compile time as opposed to run-time, which is beneficial in my opinion. Neither trait is "optional" in the sense that code requires it - code always requires a full implementation of a required trait. We can add run-time `NotImplemented` errors, but that just offloads the checks into execution as opposed to compilation in my opinion.

> Are you saying you want the PR to rename UdpStack to UdpCore? Is the TCP trait expected to similarly give up usage of the word "stack" at some point?

I think the key here is that the crate is currently so new that we aren't entirely sure what we want. I believe the consensus that we're coming to as a result of this discussion is that it would make sense to make multiple traits, for respective client and server operations for TCP and UDP, although this is still up for discussion. I personally think it would be appropriate, and it may make more sense to do a design discussion about that before trying to merge it, since it's becoming apparent here that there's a number of interested parties

src/lib.rs
```diff
@@ -65,12 +65,16 @@ pub trait UdpStack {
 	/// The type returned when we have an error
 	type Error: core::fmt::Debug;
 
-	/// Open a new UDP socket to the given address and port. UDP is connectionless,
-	/// so unlike `TcpStack` no `connect()` is required.
-	fn open(&self, remote: SocketAddr) -> Result<Self::UdpSocket, Self::Error>;
+	/// Open a UDP socket with a dynamically selected port.
```
ryan-summers, line 68: It's feasible someone may want to allocate a socket, but not connect it immediately. However, since this wasn't around in the previous implementation, I think we can ignore that use case and add it if needed later.

src/lib.rs
```diff
@@ -79,7 +83,7 @@ pub trait UdpStack {
 	/// If a packet has not been received when called, then
 	/// [`nb::Error::WouldBlock`](https://docs.rs/nb/1.0.0/nb/enum.Error.html#variant.WouldBlock)
 	/// should be returned.
-	fn read(
+	fn receive(
```
ryan-summers, line 86: Nitpick: Berkley sockets specifies this as `recv()` (and later `send_to` as `sendto`).

I personally don't care and am okay with `receive()` and `send_to()`. Just wanted to point this out.

ryan-summers (commented):
I think this is on the right path - we would also want to make these changes in TCP, but we can push that to a separate PR.

src/lib.rs
```diff
@@ -89,4 +93,21 @@ pub trait UdpStack {
 	fn close(&self, socket: Self::UdpSocket) -> Result<(), Self::Error>;
 }
 
+/// This trait is implemented by UDP/IP stacks.  It provides the ability to
+/// listen for packets on a specified port and send replies.
+pub trait UdpServerStack: UdpStack {
```
ryan-summers: I think from a clarity perspective, these traits should be `UdpClient` and `UdpServer`.

jonahd-g: I like these names the best, changed in PR.

src/lib.rs
jonahd-g: I considered that, but I wanted to stick with the more common Rust name style

jonahd-g:
The PR is updated to reflect the feedback.  I really like where it is now, and I have a [working W5500](https://github.com/jonahd-g/w5500/blob/fb5237e02639d3109888721ae81634c4494bb9c3/src/udp.rs#L194) implementation on my machine.

> I personally think it would be appropriate, and it may make more sense to do a design discussion about that before trying to merge it, since it's becoming apparent here that there's a number of interested parties

@ryan-summers how should that design discussion happen?

src/lib.rs
```diff
@@ -59,18 +59,22 @@ pub trait TcpStack {
...
+	///
+	/// Selects a port number automatically and initializes for read/writing.
+	fn connect(&mut self, remote: SocketAddr) -> Result<Self::UdpSocket, Self::Error>;
 
 	/// Send a datagram to the remote host.
-	fn write(&self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error>;
+	///
+	/// The remote host used is either the one specified in `UdpStack::connect`
+	/// or the last one used in `UdpServerStack::write_to`.
+	fn send(&mut self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error>;
```
ryan-summers: There [was a reason](https://github.com/rust-embedded-community/embedded-nal/issues/4) that none of these used a mutable reference

ryan-summers:
> @ryan-summers how should that design discussion happen?

I think it would have been beneficial to have this discussion in the issue first, but I think we came to a good conclusion here - I like where the PR is at as well and think it will address a lot of our use cases.

I also have a W5500 implementation using the 0.1.0 release of the `embedded-nal` - it would be nice to potentially add implementations of the NAL to a common place as well so that everyone can leverage the same code, but that's a future problem

jonahd-g:
> I also have a W5500 implementation using the 0.1.0 release of the embedded-nal - it would be nice to potentially add implementations of the NAL to a common place as well so that everyone can leverage the same code, but that's a future problem

I agree, see my comment on your PR :smile: \
https://github.com/kellerkindt/w5500/pull/24#issuecomment-723673384

To be clear, I've only implemented UDP, not TCP.

src/lib.rs
jonahd-g: Hmm, I'm not sure how to get my impl to work without that.  Here's my send function for example:

```rust
    fn send(&mut self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error> {
        socket.send(&mut self.bus, buffer)?;
        Ok(())
    }
```

Is there a way to internalize the need for a mutable borrow of the `bus`?

jonahd-g: nvm, found an article about it. https://ricardomartins.cc/2016/06/08/interior-mutability

jonahd-g:
Reverted the mutability change. I'll need to re-work my impl a bit, but that shouldn't block this.

MathiasKoch:
This looks great to me! Good job 👍

MathiasKoch (approved):

jonahd-g:
I have a UDP implementation of these traits for W5500 working and ready to go.  As soon as this is merged and released, I can open a PR for the driver into https://github.com/kellerkindt/w5500/

ryan-summers (approved):
Nice job! I think we got a nice API out of this. I'll follow-up with the TCP side of things here. Thanks for taking the time to get this implemented and working with us, @jonahd-g :)

eldruin (approved):

eldruin:
Thank you everybody!

PR #22, opened 2020-11-11
Added a SocketAddr struct to the return signature of Udp.receive

jonahd-g:
With the current function signature, a program can receive a packet, but has no idea of where it came from.  This alter the interface to expose that information.


ryan-summers (changes requested):
This follows from standard UDP communication, so I think it's a good change. Users who do not care about the source socket address can simply ignore the value.

TODO:
- [x] Add a changelog entry for the new return value

jonahd-g:
- [x] Add a changelog entry for the new return value

ryan-summers (approved):

MathiasKoch (approved):

eldruin (approved):

PR #23, opened 2020-11-12
Added docmentation on what read() should do if there is no packet available

jonahd-g:


ryan-summers (approved):
LGTM.

I don't think we need a changelog entry for this, since it's a documentation fix.

PR #26, opened 2020-11-16
Removed Mode enum

### LINKED ISSUE #24: [Question] Clarity on what Mode affects
I'd like some additional clarity on what Mode does, and how it affects the behavior of both read and write.  In a Write operation, there are 2 possible, and several branching ways of handling it.

1. Write, but device is busy and write operation is not possible
  a. return WouldBlock
  b. block (w/ optional timeout) until device is no-longer busy and write is possible
2. Write can be done and executes
  a. execute command and return early, assuming write will be successful
  b. block (w/ optional timeout) until write is complete and successful

For Read, there are more scenarios

1. Read, but no packets have been received
  a. return WouldBlock
  b. block (w/ optional timeout) until a packet arrives
2. Read and a packet is available
  a. read the packet into the buffer

Which of these decisions are fixed in the spec, and which are controlled by the Mode enum?  Note that there is no meaningful concept of "returning early" for the read operation.

jonahd-g:
/resolves #24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Changed
 - Changed the `UdpStack::receive` method to return the packet sender address, along with the packet length
+### Removed
+- Removed `Mode` enum, implementations should instead use `nb::WouldBlock`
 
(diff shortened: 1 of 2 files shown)


MathiasKoch (approved):

ryan-summers (approved):

eldruin (approved):
Sounds good to me, thanks!

PR #30, opened 2020-11-22
Adding initial TCP server traits

### LINKED ISSUE #10: Support for asynchronous TCP connections
The current `TcpStack` trait doesn't facilitate well for a `connect()` operation that cannot be completed in a blocking manner. For example, a socket may have called `connect()`, but `is_connected()` may return false because the TCP handshake has not completed.

As such, it's required that an application repeatedly call `connect()` on a socket whenever `is_connected()` returns false to ensure that an active connection is being made. A single call to `connect()` may fail due to TCP state mismatch and other related problems within the network stack (see https://github.com/smoltcp-rs/smoltcp/pull/367).

If the network stack isn't implemented to support repeatedly calling `connect()` on a socket that may already be connecting (but not yet connected), this can have unexpected consequences.

Is it possible for us to extend the `TcpStack` trait to either:
1. include an `is_connecting()` API
2. Modify the `is_connected()` API to return an enum indicating the state of the TCP connection (e.g. connecting, disconnected, or connected)
3. Have `connect()` return a non-exhaustive error enumeration that allows us to specify if a socket is already in the connecting process

I'm curious to hear some other thoughts on this as well. I'm looking to use these traits for an embedded MQTT client library that's targeting multiple network stacks (e.g. an external "hardware" network stack on a W5500 and a `smoltcp` network stack for a different device).

I've tried to describe the problem I'm having in an [issue for minimq](https://github.com/quartiq/minimq/issues/10), which relies on a TCP network stack, so there's further context attached there.

### LINKED ISSUE #16: Confusing TCP connection states
The current life of a TCP socket looks superfluously complex: It can be created ("open", although it's not really clear what is opened there), but only when it is connected it becomes useful. It can be closed again, but it is not clear from the documentation whether anything can be done with a closed socket other than being dropped. (Can a closed socket be connected again? From the types it could, but that wouldn't work on POSIX sockets AFAIK).

Wouldn't it be more straightforward to have a single awaitable (OK, nb::Result) client connecting function?

When TCP server functionality is added, that'd probably be centered around different types anyway (possibly a TcpServerSocket that has an accept method).

### LINKED ISSUE #27: Add support for TCP server APIs

ryan-summers:
This PR refactors the TCP traits to align with #21.

This PR fixes #27 by adding a new `TcpServer` trait.

This PR fixes #16 by clarifying the life cycle of a TCP socket. (`socket()` -> `connect()` -> `close()`).

This PR fixes #10 by updating `connect()` to return a `nb::Result` so that connect can be completed asynchronously.

**TODO**:
- [x] Implement a test use case of a TCP server using `std::net` for demonstration purposes.

CC @MathiasKoch @chrysn

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,8 +40,9 @@ pub trait TcpClient {
 	/// Check if this socket is connected
 	fn is_connected(&self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
 
-	/// Write to the stream. Returns the number of bytes written is returned
-	/// (which may be less than `buffer.len()`), or an error.
+	/// Write to the stream.
+	///
+	/// Returns the number of bytes written (which may be less than `buffer.len()`) or an error.
 	fn send(&self, socket: &mut Self::TcpSocket, buffer: &[u8]) -> nb::Result<usize, Self::Error>;
 
 	/// Receive data from the stream.
@@ -66,9 +67,9 @@ pub trait TcpClient {
 pub trait TcpServer: TcpClient {
 	/// Create a new TCP socket and bind it to the specified local port.
 	///
-	///Returns `Ok(socket)` when a new socket is successfully created and bound to the specified
-	///local port. Otherwise, an `Err(e)` variant is returned.
-	fn bind(&self, local_port: u16) -> Result<Self::TcpSocket, Self::Error>;
+	/// Returns `Ok` when a socket is successfully bound to the specified local port. Otherwise, an
+	/// `Err(e)` variant is returned.
+	fn bind(&self, socket: &mut Self::TcpSocket, local_port: u16) -> Result<(), Self::Error>;
 
 	/// Begin listening for connection requests on a previously-bound socket.
 	///
@@ -94,10 +95,13 @@ pub trait UdpClient {
 	/// The type returned when we have an error
 	type Error: core::fmt::Debug;
 
-	/// Open a UDP socket with a dynamically selected port.
+	/// Allocate a socket for further use.
+	fn socket(&self) -> Result<Self::UdpSocket, Self::Error>;
+
+	/// Connect a UDP socket with a peer using a dynamically selected port.
 	///
 	/// Selects a port number automatically and initializes for read/writing.
-	fn connect(&self, remote: SocketAddr) -> Result<Self::UdpSocket, Self::Error>;
+	fn connect(&self, socket: &mut Self::UdpSocket, remote: SocketAddr) -> Result<(), Self::Error>;
 
 	/// Send a datagram to the remote host.
 	///
@@ -124,10 +128,8 @@ pub trait UdpClient {
 /// This trait is implemented by UDP/IP stacks.  It provides the ability to
 /// listen for packets on a specified port and send replies.
 pub trait UdpServer: UdpClient {
-	/// Open a new UDP socket with a specified port
-	///
-	/// Opens a new socket with the specified port number to the given address.
-	fn bind(&self, local_port: u16) -> Result<Self::UdpSocket, Self::Error>;
+	/// Bind a UDP socket with a specified port
+	fn bind(&self, socket: &mut Self::UdpSocket, local_port: u16) -> Result<(), Self::Error>;
 
 	/// Send a packet to a remote host/port.
 	fn send_to(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,36 +14,43 @@ pub use no_std_net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, Socke
 /// which knows how to send AT commands to an ESP8266 WiFi module. You could have another implementation
 /// which knows how to driver the Rust Standard Library's `std::net` module. Given this trait, you can how
 /// write a portable HTTP client which can work with either implementation.
-pub trait TcpStack {
+pub trait TcpClient {
 	/// The type returned when we create a new TCP socket
 	type TcpSocket;
 	/// The type returned when we have an error
 	type Error: core::fmt::Debug;
 
-	/// Open a new TCP socket. The socket starts in the unconnected state.
-	fn open(&self) -> Result<Self::TcpSocket, Self::Error>;
+	/// Open a socket for usage as a TCP client.
+	///
+	/// The socket must be connected before it can be used.
+	///
+	/// Returns `Ok(socket)` if the socket was successfully created.
+	fn socket(&self) -> Result<Self::TcpSocket, Self::Error>;
 
 	/// Connect to the given remote host and port.
+	///
+	/// Returns `Ok` if the connection was successful. Otherwise, if the connection could not be
+	/// completed immediately, this function should return [`nb::Error::WouldBlock`].
 	fn connect(
 		&self,
-		socket: Self::TcpSocket,
+		socket: &mut Self::TcpSocket,
 		remote: SocketAddr,
-	) -> Result<Self::TcpSocket, Self::Error>;
+	) -> nb::Result<(), Self::Error>;
 
 	/// Check if this socket is connected
 	fn is_connected(&self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
 
 	/// Write to the stream. Returns the number of bytes written is returned
 	/// (which may be less than `buffer.len()`), or an error.
-	fn write(&self, socket: &mut Self::TcpSocket, buffer: &[u8]) -> nb::Result<usize, Self::Error>;
+	fn send(&self, socket: &mut Self::TcpSocket, buffer: &[u8]) -> nb::Result<usize, Self::Error>;
 
-	/// Read from the stream.
+	/// Receive data from the stream.
 	///
 	/// Returns `Ok(n)`, which means `n` bytes of data have been received and
 	/// they have been placed in `&buffer[0..n]`, or an error. If a packet has
 	/// not been received when called, then [`nb::Error::WouldBlock`]
 	/// should be returned.
-	fn read(
+	fn receive(
 		&self,
 		socket: &mut Self::TcpSocket,
 		buffer: &mut [u8],
@@ -51,6 +58,29 @@ pub trait TcpStack {
 
 	/// Close an existing TCP socket.
 	fn close(&self, socket: Self::TcpSocket) -> Result<(), Self::Error>;
+}
+
+/// This trait is implemented by TCP/IP stacks that expose TCP server functionality. TCP servers
+/// may listen for connection requests to establish multiple unique TCP connections with various
+/// clients.
+pub trait TcpServer: TcpClient {
+	/// Create a new TCP socket and bind it to the specified local port.
+	///
+	///Returns `Ok(socket)` when a new socket is successfully created and bound to the specified
+	///local port. Otherwise, an `Err(e)` variant is returned.
+	fn bind(&self, local_port: u16) -> Result<Self::TcpSocket, Self::Error>;
+
+	/// Begin listening for connection requests on a previously-bound socket.
+	///
+	/// Returns `Ok` if the socket was successfully transitioned to the listening state. Otherwise,
+	/// an `Err(e)` variant is returned.
+	fn listen(&self, socket: &mut Self::TcpSocket) -> Result<(), Self::Error>;
+
+	/// Accept an active connection request on a listening socket.
+	///
+	/// Returns `Ok(connection)` if a new connection was created. If no pending connections are
+	/// available, this function should return [`nb::Error::WouldBlock`].
+	fn accept(&self, socket: &mut Self::TcpSocket) -> nb::Result<Self::TcpSocket, Self::Error>;
 }
 
 /// This trait is implemented by UDP/IP stacks. You could, for example, have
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -14,36 +14,43 @@ pub use no_std_net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, Socke
...
 	type Error: core::fmt::Debug;
 
-	/// Open a new TCP socket. The socket starts in the unconnected state.
-	fn open(&self) -> Result<Self::TcpSocket, Self::Error>;
+	/// Open a socket for usage as a TCP client.
+	///
+	/// The socket must be connected before it can be used.
+	///
+	/// Returns `Ok(socket)` if the socket was successfully created.
+	fn socket(&self) -> Result<Self::TcpSocket, Self::Error>;
```
jonahd-g, line 28: I wonder if the UdpClient trait should have a `socket` method as well for consistency?

src/lib.rs
```diff
@@ -14,36 +14,43 @@ pub use no_std_net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, Socke
...
 	fn connect(
 		&self,
-		socket: Self::TcpSocket,
+		socket: &mut Self::TcpSocket,
 		remote: SocketAddr,
-	) -> Result<Self::TcpSocket, Self::Error>;
+	) -> nb::Result<(), Self::Error>;
 
 	/// Check if this socket is connected
 	fn is_connected(&self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
```
jonahd-g, line 41: Should this function stay even with the new `nb::Result` return of `connect`?

src/lib.rs
```diff
@@ -53,6 +60,29 @@ pub trait TcpStack {
...
+	///
+	///Returns `Ok(socket)` when a new socket is successfully created and bound to the specified
+	///local port. Otherwise, an `Err(e)` variant is returned.
+	fn bind(&self, local_port: u16) -> Result<Self::TcpSocket, Self::Error>;
+
+	/// Begin listening for connection requests on a previously-bound socket.
+	///
+	/// Returns `Ok` if the socket was successfully transitioned to the listening state. Otherwise,
+	/// an `Err(e)` variant is returned.
+	fn listen(&self, socket: &mut Self::TcpSocket) -> Result<(), Self::Error>;
```
jonahd-g, line 78: What is the value in this extra `listen` step?  What drawbacks would there be in having `bind` go into the listening state?

jonahd-g (commented):
I don't know much about the TCP side of things.  Is it possible for a single socket to communicate with multiple clients at once?  Or would you need multiple sockets listening on the same port?  Or something else?

src/lib.rs
ryan-summers: I would imagine that checking if a socket is currently connected does not require any delay (it would typically be a status register check, software boolean state check, or single register read). As such, I don't think it warrants an `nb::Result` since the check wouldn't be a long operation - open to other use cases here that would warrant an `nb::Result` though!

src/lib.rs
ryan-summers: I think your motivation here is right - we should have a symmetric API for both UDP and TCP. 

Should we add a `socket()` method to UDP for consistency? I don't think it would be a bad idea. My only qualm is that it doesn't make much sense to have a UDP socket in an un-opened state. What does `UdpClient::socket()` do that `UdpClient::connect()` does not do?

---

An aside about why this design is necessary for the TCP client:

To explain my design here, an alternative question is as follows:
* What's the purpose in not returning a `TcpSocket` in the result of `connect()`?

I went with this approach because `connect()` is asynchronous (requires a 3-way handshake and lots of communication between client and server before connection is established), so it has to return an `nb::Result`. What happens if `connect()` needs to return `nb::WouldBlock`? In that case, it can't return the `TcpSocket`, so there's no object returned to track the state of the connection request. Should future calls to `connect()` assume state from previous calls? That seems incorrect, which is why I went with a `socket()` constructor here.

ryan-summers:
@jonahd-g 
> I don't know much about the TCP side of things. Is it possible for a single socket to communicate with multiple clients at once?

With a TCP server, you typically have a single socket listening on a given local port. It then receives connection requests from various remote clients and creates a new connection for each request. Each connection is assigned a unique socket for the communication stream, but the original listening socket remains operational to facilitate more connections. Thus, a single TCP server can serve any number of TCP clients that it wants to support simultaneously.

> Or would you need multiple sockets listening on the same port? Or something else?

Based on my reading of [`berkley-sockets`](https://en.wikipedia.org/wiki/Berkeley_sockets):
> _accept() is used on the server side. It accepts a received incoming attempt to create a new TCP connection from the remote client, and creates a new socket associated with the socket address pair of this connection._

There is a single listening socket that creates new sockets for each accepted connection.

src/lib.rs
ryan-summers: I would imagine that a user would like the ability to wait until a device is powered and operational (e.g. after all initialization) before starting to listen and accept incoming connections.

In my use cases, I'm interested in directly `listen()`-ing the socket, but I could imagine cases where a user does _not_ want that. This API is following the `berkley-sockets` model as well.

src/lib.rs
jonahd-g: `UdpClient:connect` configures the device with the remote address, that's all.

src/lib.rs
jonahd-g: My question was worded confusingly.  I mean, do we need this function at all?  Is there a reason to check if a socket is connected?

jonahd-g:
> There is a single listening socket that creates new sockets for each accepted connection.

That makes sense, thanks for explaining.

src/lib.rs
jonahd-g: > What does UdpClient::socket() do that UdpClient::connect() does not do?

socket() would allocate one of a potentially finite number of available sockets.  Right now, the trait has that action wrapped up into `bind` or `connect`

src/lib.rs
```diff
@@ -14,36 +14,43 @@ pub use no_std_net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, Socke
...
-		socket: Self::TcpSocket,
+		socket: &mut Self::TcpSocket,
 		remote: SocketAddr,
-	) -> Result<Self::TcpSocket, Self::Error>;
+	) -> nb::Result<(), Self::Error>;
 
 	/// Check if this socket is connected
 	fn is_connected(&self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
 
 	/// Write to the stream. Returns the number of bytes written is returned
```
jonahd-g: "return" used twice.  Maybe

> ```
> Write to the stream.
>
> Returns the number of bytes written (which may be less than `buffer.len()`), or an error.

src/lib.rs
```diff
@@ -53,6 +60,29 @@ pub trait TcpStack {
...
 
+/// This trait is implemented by TCP/IP stacks that expose TCP server functionality. TCP servers
+/// may listen for connection requests to establish multiple unique TCP connections with various
+/// clients.
+pub trait TcpServer: TcpClient {
+	/// Create a new TCP socket and bind it to the specified local port.
+	///
+	///Returns `Ok(socket)` when a new socket is successfully created and bound to the specified
+	///local port. Otherwise, an `Err(e)` variant is returned.
+	fn bind(&self, local_port: u16) -> Result<Self::TcpSocket, Self::Error>;
```
jonahd-g: Why does `bind()` create and return a socket, but `connect()` _accepts_ a socket?  Is that intentional?

src/lib.rs
```diff
@@ -53,6 +60,29 @@ pub trait TcpStack {
...
+	/// Begin listening for connection requests on a previously-bound socket.
+	///
+	/// Returns `Ok` if the socket was successfully transitioned to the listening state. Otherwise,
+	/// an `Err(e)` variant is returned.
+	fn listen(&self, socket: &mut Self::TcpSocket) -> Result<(), Self::Error>;
+
+	/// Accept an active connection request on a listening socket.
+	///
+	/// Returns `Ok(connection)` if a new connection was created. If no pending connections are
+	/// available, this function should return [`nb::Error::WouldBlock`].
```
jonahd-g, line 83: Should also return `WouldBlock` if there are no available sockets to use.  e.g. W5500 only has 8, if they're all busy then the connection cannot be accepted.

jonahd-g (changes requested):

src/lib.rs
```diff
@@ -53,6 +60,29 @@ pub trait TcpStack {
...
 }
 
+/// This trait is implemented by TCP/IP stacks that expose TCP server functionality. TCP servers
+/// may listen for connection requests to establish multiple unique TCP connections with various
+/// clients.
+pub trait TcpServer: TcpClient {
+	/// Create a new TCP socket and bind it to the specified local port.
+	///
+	///Returns `Ok(socket)` when a new socket is successfully created and bound to the specified
+	///local port. Otherwise, an `Err(e)` variant is returned.
```
MathiasKoch:
```suggestion
	/// Returns `Ok(socket)` when a new socket is successfully created and bound to the specified
	/// local port. Otherwise, an `Err(e)` variant is returned.
```

MathiasKoch (approved):
LGTM, once the stuff from @jonahd-g has been resolved :+1: 

Great work!

src/lib.rs
Sympatron: If the `socket()` call only allocates the required resources, then `connect()` is definitely necessary to check whether the handshake is still in progress/done or if it failed (e.g. timeout).

src/lib.rs
ryan-summers: > `UdpClient:connect` configures the device with the remote address, that's all.

Ah, I totally forgot that UDP is connection-less, so there's no data transfer and no logical "stream" between devices. It's essentially shouting into the void - that makes sense, thanks! I think it's appropriate to make a symmetric `socket()` function for UDP as well.

src/lib.rs
ryan-summers: It's also possible for the remote peer to close a connection during communication, so we need to be able to indicate that through the stack as well. Typically, the user should not be `send()`ing or `receive()`ing on the socket if it is not connected, as that would generate an error.

src/lib.rs
ryan-summers: I would imagine that this should actually be an `Error` returned by the network stack instead. Simply waiting longer would not necessarily result in a free socket in the future as there's no guarantee that one of the other users will ever release their socket.

src/lib.rs
ryan-summers: I have updated all sockets to be created using `socket()` and then updated all other functions to accept a reference to the original socket.

src/lib.rs
ryan-summers: I have updated all sockets to be created using socket() and then updated all other functions to accept a reference to the original socket.

src/lib.rs
jonahd-g: That makes sense

ryan-summers:
I have now written a small TCP echo server using the traits to verify that everything works as I would expect to - there was one modification to update `accept()` to return the connected client's address, but other than that, everything worked nicely. This should be ready for merge I believe @MathiasKoch, @Sympatron and @jonahd-g.

If we'd like to add type-state control of the TCP listening socket, I would advocate we wait for another PR. There's some difficulties because of the `open()` and `close()` APIs having to get re-implemented.

MathiasKoch:
> If we'd like to add type-state control of the TCP listening socket, I would advocate we wait for another PR. There's some difficulties because of the open() and close() APIs having to get re-implemented.

I agree on this, as it seems like a super nice to have approach, but kind of out-of-reach at this point, unless one of you guys have the time to draft up such an API?

So from my side of things, LGTM

jonahd-g:
I'm willing to start working on a typestate solution, but I'd really like to see this current approach merged and released in the mean-time, considering the currently-released version does not support binding.

Sympatron:
LGTM. I think we can merge it the way it is and add type states in a new PR with a separate discussion.

ryan-summers:
CC: @eldruin - this should now be ready for merge

eldruin:
Thanks @ryan-summers! GitHub complains about conflicts, though. Could you rebase this?

ryan-summers:
Alright, should now be fixed up after a rebase (there was indeed a conflict)

eldruin:
Thanks everyone!

PR #34, opened 2020-12-02
Release 0.2.0

eldruin:
I added @ryan-summers as author as well. Is that email address fine for you?

cc. @MathiasKoch @ryan-summers


ryan-summers (approved):
LGTM

eldruin:
This is now published on crates.io and tagged here.

Issue #31, opened 2020-11-24
UdpClient/UdpServer may be confusing to users


Sympatron:
Looking at the whole `UdpClient`/`UdpServer` thing from #21 I think the idea behind the trait it kind of changed.

Before it was called `UdpStack` and represented the whole network interface that creates sockets on demand for new connections, but now the same suddenly a `UdpServer` that creates either `UdpSockets` in server or client mode depending on the function you call. I think it is kind of confusing that your network interface has to implement `UdpServer` now.
I have no definitive answer how to make the API clearer, but maybe the socket types should be different in client and server mode instead of the stack. The stack traits could be called `UdpStack`/`UdpServerStack` or something like `UdpSimpleStack`/`UdpAdvancedStack` or `UdpStackCore`/... Maybe somebody has a good idea for names.

The point I am trying to make is, that it may be confusing to mix semantics of network interfaces and their traits and actual sockets. Similar things apply to TCP (#30) of course.

ryan-summers:
To my knowledge, a UDP socket is the same (a simple data pipe with RX/TX) regardless of client/server roles, so there's no point in differentiating between a socket created by a UDP server and a socket created by a UDP client once it's connected to the peer.

It's a bit of a misnomer that we had `UdpStack` previously because it _wasn't actually a stack_. Instead, it only exposed client-specific APIs and had no support for the server role at all. If we wanted to bring back `UdpStack`, it would likely be exactly what is currently `UdpServer`.

Sympatron:
What I meant was that the `UdpStack`/`UdpServer` trait represents the whole network interface, or am I getting something wrong here? The connection is represented by `UdpSocket`, but `UdpStack` was basically supposed to be a "global" entity, or not?
Therefore the current naming scheme is confusing, because `UdpClient`/`UdpServer` implies a connection in my opinion.

With UDP you are correct, that server and client are quite similar, but we could still represent the current mode with the type model to ensure proper usage at compile time. At the moment you could call `send()` on a server socket which doesn't make sense, because then you can't specify the recipient. And you can call `send_to()` on a client that is already connected to a (potentially different) remote host. These errors can easily be detected at compile time with a stricter type system.

ryan-summers:
> What I meant was that the `UdpStack`/`UdpServer` trait represents the whole network interface, or am I getting something wrong here? The connection is represented by `UdpSocket`, but `UdpStack` was basically supposed to be a "global" entity, or not?
[...]

PR #36, opened 2020-12-03
Renamed UDP/TCP traits to alleviate confusion

Sympatron:
Based on our discussion in #31 I renamed the traits to make their intent more clear.


eldruin (changes requested):
Thanks @Sympatron. Could you add an entry to the Changelog as well?

ryan-summers (approved):
Outside of the changelog, this LGTM

eldruin (approved):
Sounds good to me, thanks!

PR #37, opened 2020-12-03
Made Dns method names more rusty

Sympatron:
This is a minor nitpick, but I think it would be better to use idiomatic Rust names for things, than legacy C names.
And it would be a convenient time, because we are introducing breaking changes with #36 anyway.


MathiasKoch (approved):
LGTM :+1:

eldruin (approved):
Sounds good to me, thanks!

PR #42, opened 2021-01-17
Removed Internal Mutability

### LINKED ISSUE #39: Revisit on mutable self references
I don't understand the reason for immutable self references that clearly do mutate the state of self or require exclusive access to the underlying implementation or bus. I think this design decision only causes [hacks to circumvent it](https://gitlab.com/etonomy/riot-wrappers/-/blob/master/src/socket_embedded_nal.rs#L101), unnecessary overhead in the [driver implementation](https://github.com/kellerkindt/w5500/blob/29cc295010c526237d09f0f0b590343fadc324f3/src/interface.rs#L13) (which is the opposite of 'zero cost'/'you only pay for what you need') and does not reflect the strategy of the rust ecosystem, such as  [embedded-hal](https://github.com/rust-embedded/embedded-hal/blob/eae6c9958f8af4a3d0e02b31238efeb2c80e4e15/src/spi.rs#L21), [std-Read](https://doc.rust-lang.org/std/io/trait.Read.html), [std-Write](https://doc.rust-lang.org/std/io/trait.Write.html), [tokio-AsyncRead](https://docs.rs/tokio/1.0.1/tokio/io/trait.AsyncWrite.html) or [tokio-AsyncWrite](https://docs.rs/tokio/1.0.1/tokio/io/trait.AsyncWrite.html).

I don't follow the argument of @thejpster [here](https://github.com/rust-embedded-community/embedded-nal/issues/4#issuecomment-589332989) as it seems more of an special case scenario and could be solved otherwise by implementing `Clone` on what seems to be an FD or raw-pointer(?) internally.

Feel free to convince me otherwise.

We stumbled over this in a [discussion here](https://github.com/kellerkindt/w5500/pull/26) where @jonahd-g (thanky you!) proposes an implementation of embedded-nal for the w5500 driver.

jonahd-g:
Removed requirement on internal mutability by using mutable self references in methods.

See #39 for details.  Should not be merged unless that alternate approach to sharing is approved.

/resolves #39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Changed the names of `UdpClient`/`TcpClient` to `UdpClientStack`/`TcpClientStack`
 - Changed the names of `UdpServer`/`TcpServer` to `UdpFullStack`/`TcpFullStack`
 - Changed the method names `Dns::gethostbyname`/`Dns::gethostbyaddr` to `Dns::get_host_by_name`/`Dns::get_host_by_address`
+- Changed self references in all network stack methods to mutable, with the intent of handling sharing in a different layer (see [#43](https://github.com/rust-embedded-community/embedded-nal/issues/43)).
 
 ## [0.2.0] - 2020-12-02
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,25 +26,29 @@ pub trait TcpClientStack {
 	/// The socket must be connected before it can be used.
 	///
 	/// Returns `Ok(socket)` if the socket was successfully created.
-	fn socket(&self) -> Result<Self::TcpSocket, Self::Error>;
+	fn socket(&mut self) -> Result<Self::TcpSocket, Self::Error>;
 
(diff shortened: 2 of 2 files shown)


ryan-summers:
I agree with your point about internal mutability being required and the existing API not guaranteeing thread safety, and thus, I am on board with this. @MathiasKoch thoughts?

MathiasKoch:
I think this is okay, assuming we can reach some agreement on #39 as stated in OP

thejpster (approved):

ryan-summers (approved):

ryan-summers (changes requested):
Ah just noticed missing CHANGELOG entry. Mind adding a notice there that we're adding self mutability and reference issue #39 for more detail?

eldruin (approved):
I agree with this way forward.
Other than the changelog entry, looks good to me. Thanks!

MathiasKoch (approved):

ryan-summers (approved):
Thanks! :)

jonahd-g:
Thanks Ryan.  I keep forgetting too, maybe we need an Action to check for that.

Changelog entry added.

Issue #43, opened 2021-01-18
Shared NAL


jonahbron:
Once #42 is merged, the NAL traits in this crate will no-longer be sharable on their own.  It is necessary to create a new layer that can handle the sharing internally.

This can be split into two different types of sharing: single-thread (where RefCell can be used), and multi-threaded.  Both can draw inspiration from [Rahix/shared-bus](/Rahix/shared-bus).

I've drafted a simple RefCell-based wrapper here, inspired by the shared-bus Simple Manager.  It's suitable for sharing access to the network stack in a single-threaded context.  However @ryan-summers you mentioned that you think there may be issues with this approach?  Would you mind clarifying?

More work will be necessary to create a sharing layer that's either Sync or Send.

```rust
use core::cell::RefCell;
use embedded_nal::{nb, SocketAddr, TcpClientStack, TcpFullStack, UdpClientStack, UdpFullStack};

pub struct SharedNalManager<T> {
    driver: RefCell<T>,
}

impl<T> SharedNalManager<T> {
    pub fn new(driver: T) -> Self {
        SharedNalManager {
            driver: RefCell::new(driver),
        }
    }

    pub fn acquire(&self) -> SharedNal<T> {
        SharedNal {
            driver: &self.driver,
        }
    }
}

pub struct SharedNal<'a, T> {
    driver: &'a RefCell<T>,
}

macro_rules! forward {
    ($func:ident($($v:ident: $IT:ty),*) -> $T:ty) => {
        fn $func(&mut self, $($v: $IT),*) -> $T {
            self.driver.borrow_mut().$func($($v),*)
        }
    }
}

impl<'a, T: UdpClientStack> UdpClientStack for SharedNal<'a, T> {
    type Error = T::Error;
    type UdpSocket = T::UdpSocket;

    forward! {socket() -> Result<Self::UdpSocket, Self::Error>}
    forward! {connect(socket: &mut Self::UdpSocket, address: SocketAddr) -> Result<(), Self::Error>}
    forward! {send(socket: &mut Self::UdpSocket, data: &[u8]) -> Result<(), nb::Error<<T as embedded_nal::UdpClientStack>::Error>>}
    forward! {receive(socket: &mut Self::UdpSocket, data: &mut [u8]) -> Result<(usize, SocketAddr), nb::Error<<T as UdpClientStack>::Error>>}
    forward! {close(socket: Self::UdpSocket) -> Result<(), Self::Error>}
}

[...]

PR #45, opened 2021-01-20
Moved network stack traits into a module

jonahbron:
Moved network stack traits into a module, leaving behind only the public interface definition in lib.rs

This is in preparation to add a `src/stack/share.rs` module for the new sharing layer.  The public interface is entirely unchanged.  Lmk if anybody dislikes this change, I just think it's a little tidier to use lib.rs to define the public interface.

You can verify that I didn't mess up anything in the move by going to the Commits tab and selecting the 
[first commit](https://github.com/rust-embedded-community/embedded-nal/pull/45/commits/7d0e6874cf9fc2b75850afd02054e6a252bee386).

#43


src/stack/mod.rs
```diff
@@ -0,0 +1,142 @@
+use no_std_net::SocketAddr;
```
ryan-summers: What are your thoughts about potentially even splitting this into a `udp.rs` and `tcp.rs` file as well? In general, I like this cleanup change otherwise.

jonahbron: I'm in favor of that actually, but didn't want to shake things up too much and risk push-back :)

ryan-summers: Let's go for that. I think it's quite a logical representation - agreed that `lib.rs` should be for the public API only

jonahbron: Split performed

MathiasKoch (approved):
Lgtm

jonahbron:
I have the sharing layer code in a branch based off of this one ready to go.  Once this is approved and merged, I'll open another PR for it.

ryan-summers (approved):

PR #46, opened 2021-01-21
Added a shareable wrapper for NAL stack implementations

### LINKED ISSUE #43: Shared NAL

jonahbron:
Ended up with some slightly different names that made more sense to me in the context of the traits and docs.  Open to feedback.

/fixes #43

### DIFF
--- a/src/stack/share.rs
+++ b/src/stack/share.rs
@@ -1,0 +1,142 @@
+use crate::{nb, SocketAddr, TcpClientStack, TcpFullStack, UdpClientStack, UdpFullStack};
+use core::cell::RefCell;
+
+/// Sharable wrapper for a network stack implementation.
+///
+/// An implementation of the stack traits that can contain (and provide provide
+/// single-threaded shared access to) another stack implementation. A direct
+/// implementation can only be used when owned or with a mutable reference.
+/// This implementation will store another implementation internally, and yield
+/// an arbitrary number of shared references to it, which themselves implement
+/// the stack traits.
+///
+/// ```
+/// use embedded_nal::SharableStack;
+/// # use embedded_nal::{UdpClientStack, SocketAddr, SocketAddrV4, Ipv4Addr, nb};
+/// # struct SomeNalDriver {}
+/// # impl SomeNalDriver {
+/// #   fn new() -> Self { Self {} }
+/// # }
+/// # impl UdpClientStack for SomeNalDriver {
+/// #   type Error = ();
+/// #   type UdpSocket = ();
+/// #   fn socket(&mut self) -> Result<Self::UdpSocket, Self::Error> {
+/// #     Ok(())
+/// #   }
+/// #   fn connect(
+/// #   	&mut self,
+/// #   	socket: &mut Self::UdpSocket,
+/// #   	remote: SocketAddr,
+/// #   ) -> Result<(), Self::Error> {
+/// #     Ok(())
+/// #   }
+/// #   fn send(&mut self, socket: &mut Self::UdpSocket, buffer: &[u8]) -> nb::Result<(), Self::Error> {
+/// #     Ok(())
+/// #   }
+/// #   fn receive(
+/// #   	&mut self,
+/// #   	socket: &mut Self::UdpSocket,
+/// #   	buffer: &mut [u8],
+/// #   ) -> nb::Result<(usize, SocketAddr), Self::Error> {
+/// #     Ok((0, SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 0))))
+/// #   }
+/// #   fn close(&mut self, socket: Self::UdpSocket) -> Result<(), Self::Error> {
+/// #     Ok(())
+/// #   }
+/// # }
+/// let mut driver = SomeNalDriver::new();
+/// // Driver can only be used in one place at a time.
+/// let mut sharable_driver = SharableStack::new(driver);
+/// // Sharable driver can't do anything on its own, but it can create many usable copies.
+/// let mut shared_driver0 = sharable_driver.acquire();
+/// let mut shared_driver1 = sharable_driver.acquire();
+/// // These shared copies can be passed around to other parts of an application's code, and used
+/// // independently.
+/// let mut socket0 = shared_driver0.socket()?;
+/// shared_driver0.connect(&mut socket0, SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080)));
+/// // ...
+///
+/// // ... and somewhere else
+/// let mut socket1 = shared_driver1.socket()?;
+/// shared_driver1.connect(&mut socket1, SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8443)));
+/// // ...
+/// # Ok::<(), ()>(())
+/// ```
+pub struct SharableStack<T> {
+	stack: RefCell<T>,
+}
+
+impl<T> SharableStack<T> {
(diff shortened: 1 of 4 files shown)


src/stack/share.rs
```diff
@@ -0,0 +1,77 @@
...
+
+macro_rules! forward {
+    ($func:ident($($v:ident: $IT:ty),*) -> $T:ty) => {
+        fn $func(&mut self, $($v: $IT),*) -> $T {
+            self.stack.borrow_mut().$func($($v),*)
+        }
+    }
+}
+
+impl<'a, T: UdpClientStack> UdpClientStack for SharedStack<'a, T> {
```
ryan-summers: I would recommend opting for the `where` clause instead of `T: UdpClientStack` - I find it more readable and easy to follow.

ryan-summers: This feels like something we may need to have a test for to verify that we can make multiple copies of it.

src/stack/share.rs
```diff
@@ -0,0 +1,77 @@
...
+	/// Returns a shared reference to the driver that can be used as a first-class implementation.
+	pub fn acquire(&self) -> SharedStack<T> {
+		SharedStack { stack: &self.stack }
+	}
+}
+
+/// Single-thread shared reference to an internal network stack implementation.
+///
+/// This can only be created by calling [`SharableStack::acquire()`]
+pub struct SharedStack<'a, T> {
```
ryan-summers, line 86: I think we could provide some sort of example usage here?

src/stack/share.rs
jonahbron: With the magic of doc tests, I've provided both a usage example _and_ a sharing test at once :smile:

jonahbron:
The rust doc.  This works with doc tests too.

image

jonahbron:
@Rahix Since we've been referencing your work on this, seems logical that you be able to offer feedback.

jonahbron:
Ping again for feedback on this.

ryan-summers (approved):
LGTM. Thanks for the ping and the work on this! I'm going to be prepping a new release soon as well.

ryan-summers:
@jonahbron Please add a CHANGELOG entry and I'll merge this

Rahix:
Hey!

Sorry, I didn't get around to taking a look yet and quite frankly the thread got buried in my inbox so I forgot :/

Basically there isn't much to get wrong though as long as no unsafe or multi-tasking is involved (multi-tasking is prevented by `!Send` on the `RefCell`).  So I think the implementation is solid :)

One idea I'd like to put out there (though this could always be considered later on as well):  Maybe it would make sense to add this support into `shared-bus` instead?  This way, you could benefit from all the other mutex implementations over there as well, with zero additional effort.

MathiasKoch (approved):
Only thing is if it would be worth having the option to somehow catch `borrow_mut()` before panic ? `try_borrow_mut()`? But i think it would screw up the error types quite alot.

Rahix:
@MathiasKoch, I do not think there is much to be gained.  The only reason why `borrow_mut()` could panic would be if the stack somehow accesses another `SharedStack` reference to itself.  I would argue that this is a serious programming error and not a failure condition that should be handled in production.

MathiasKoch:
Then i think all we are missing is a changelog entry from @jonahbron before we can merge this and do a new 0.3 release.

> One idea I'd like to put out there (though this could always be considered later on as well): Maybe it would make sense to add this support into shared-bus instead? This way, you could benefit from all the other mutex implementations over there as well, with zero additional effort.

I think this would be nice eventually, but might be much easier once the soundness of this design is proven?

jonahbron:
@ryan-summers changelog added

ryan-summers (approved):

PR #48, opened 2021-02-02
pub use heapless

### LINKED ISSUE #17: Re-exports of publicly used crates
The state of crates used publicly is currently inconsistent:

* nb (which is used in many error types) is `pub use`d.
* heapless (which is part of the return type of [gethostbyaddr](https://docs.rs/embedded-nal/0.1.0/embedded_nal/trait.Dns.html#tymethod.gethostbyaddr)) is not.

This means that as an implementer of embedded-nal I need to depend on heapless and hope for the versions to match. The straightforward way to mitigate would be pub use of heapless.

chrysn:
This is needed by embedded-nal trait implementers who build get_host_by_address results, or by trait users who pass the results on.

Closes: https://github.com/rust-embedded-community/embedded-nal/issues/17

Without this, they have to pull in a heapless dependency on their own and pray it resolves to the same version (aided by careful adjustment of the version they depend on). With this, they can just operate on the type embedded-nal uses.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,9 @@ mod dns;
 mod stack;
 
 pub use nb;
+// Needed by embedded-nal trait implementers who build get_host_by_address results, or by trait
+// users who pass the results on.
+pub use heapless;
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Sounds good to me, thanks!

PR #50, opened 2021-03-05
Change DNS trait to return nb, and bump dependencies

### LINKED ISSUE #35: Evaluate effectiveness of DNS traits
It is currently my understanding that DNS can be implemented on top of generic sockets. As such, it should be possible to write a generic `no_std` DNS client crate that relies on the `embedded-nal`. This would mean that the dns traits in this crate could likely be removed, since they no longer need to be generic.

I'm curious to hear other opinions here - I have not yet written a DNS client, but likely will be doing so in the near future (and would publish it as said `no_std` DNS resolver)

MathiasKoch:
- Changed `Dns` methods to return `nb::Result<..>` to allow non-blocking implementations.
- Bump dependency version of `heapless` to `v0.6.1` to address security issue of sub-dependency.
- Bump dependency version of `no-std-net` to `v0.5`.

Fixes #35

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
 
         include:
           # Test MSRV
-          - rust: 1.36.0
+          - rust: 1.46.0
             TARGET: x86_64-unknown-linux-gnu
 
           # Test nightly but don't fail
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Changed [`Dns`](./src/dns.rs) methods to return `nb::Result<..>` to allow non-blocking implementations.
 - Bump dependency version of `heapless` to `v0.6.1` to address security issue of sub-dependency.
 - Bump dependency version of `no-std-net` to `v0.5`.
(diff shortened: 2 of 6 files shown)


ryan-summers (approved):

eldruin (approved):
Looks good to me!

MathiasKoch:
@ryan-summers & @eldruin This does require bumping MSRV to 1.39.0. Any issues with that?

ryan-summers:
I think one of the dependencies may not be compatible with 1.36.0.

Edit: Bumping MSRV is fine with me

eldruin (approved):
Bumping MSRV is fine for me, thanks!

MathiasKoch:
Huh. Apparently requires 1.46 due to const_fn in loops: https://medium.com/@knoldus/whats-new-in-rust-1-46-0-about-const-fn-7d58f92d719f

@eldruin @ryan-summers  Should i bump the version aswell so we can release it right away, or do you want to wait for more stuff?

eldruin (approved):
Bump and release would be fine for me.

MathiasKoch:
@eldruin Should be ready for a publish now :+1: Thanks

eldruin:
This is now published :tada: 

@MathiasKoch You should also have publishing permission, though.

PR #51, opened 2021-03-07
tcp: Fix small doc comment typo

dnadlinger:


MathiasKoch:
Thanks for correcting 👍

PR #54, opened 2021-05-18
Update heapless to v0.7.0 and use const generics

MathiasKoch:


src/dns.rs
```diff
@@ -43,5 +43,5 @@ pub trait Dns {
 	/// 255 bytes [`rfc1035`]
 	///
 	/// [`rfc1035`]: https://tools.ietf.org/html/rfc1035
-	fn get_host_by_address(&self, addr: IpAddr) -> nb::Result<String<consts::U256>, Self::Error>;
+	fn get_host_by_address(&self, addr: IpAddr) -> nb::Result<String<256>, Self::Error>;
```
ryan-summers, line 46: Update the README and MSRV as well? Or are we already at MSRV 1.52?

MathiasKoch: Ohh, yeah missed that.. I think 1.51 should be sufficient?

ryan-summers (approved):

eldruin (approved):
Looks good to me as well. Thanks!

PR #56, opened 2021-05-21
Change self references in dns stack methods to mutable

MathiasKoch:
I think the DNS trait was overlooked, when the change to mutable self references happened for the network stacks?

This changes self references in dns stack methods to mutable, to follow the network stack implementations.


ryan-summers (approved):

eldruin (approved):
Sounds good to me as well. Thanks!

PR #57, opened 2021-05-25
Preparing v0.6.0 release

MathiasKoch:


eldruin (commented):
Looks good to me but the version bump in the Cargo.toml is missing.

MathiasKoch:
Huh, forgot to save the file ;) 

Good catch

PR #60, opened 2021-09-22
Add blanket impls for mutable references

MathiasKoch:
I recently ran into an issue where i had a closure of signature `F: FnOnce(&mut dyn TcpClientStack<TcpSocket = Handle, Error = Error>)`, which requires an impl of the `embedded-nal` traits for mutable references as `impl<T: TcpClientStack> TcpClientStack for &mut T`, in order to be used in functions like `fn foo<T: TcpClientStack>(network: &mut T) { ... }`.

This PR adds those blanket implementations for all the traits in this repo.

This is also in line with the agreed upon way in `embedded-hal`: https://github.com/rust-embedded/embedded-hal/pull/310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-* None
+* Add blanket impls of all the traits for mutable references.
 
 ## [0.6.0] - 2021-05-25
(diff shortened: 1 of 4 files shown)


eldruin (commented):
Looks good to me, thanks!

eldruin (approved):

MathiasKoch:
Ill just add a Changelog entry, and merge :+1:

Issue #63, opened 2021-11-10
TCP: Typestate client and server sockets


chrysn:
For TCP, there are two kinds of sockets, those bound (on which `accept` makes sense), and those connected or accepted (on which read/write makes sense). These are the same type in POSIX systems, but completely different eg. [in RIOT](https://doc.riot-os.org/group__net__sock__tcp.html#ga50008ef716746eb6d2a50c56615d8547).

This is similar to #33 in methods, but different in goals (#33 is about when what can be called on the same sockets; feel free to close as duplicate if you disagree).

I suggest that the Socket associated type of TcpClientStack be split into a ConnectionSocket and a ListeningSocket, which can still be the same types on POSIX-style systems but distinct on others. (If #33 is followed, a ConnectionSocket would then be further split into its unconnected and its connected version, and the listening into its unbound, bound and listening versions).

Alternatively, sockets on such platforms need to be union-typed (`enum Socket { Unspecified, ConnectionSocket(sock_t), ListeningSocket(sock_listener_t) }`), and always check on runtime whether the right kind is used, which feels not very idiomatic.

chrysn:
Some of this is now implemented in the TcpConnect API introduced in embedded-nal-async 0.2: Rather than having sockets that mutate through the connection process, the connect() method returns a future whose result is a connected socket at the type level.

PR #70, opened 2022-08-03
Remove tcp stack

lulf:
Remove TcpClientStack and TcpFullStack 
 
These traits restricts implementors to only support 1connection at a time, which only covers a small set of use cases.
 
The new TcpConnect trait is more flexible and should be the preferred trait for implementors and users. 

NOTE:

* UDP traits remain, I'm not sure if they should be removed or not. It is tempting to remove it so that we can start with a clean slate now before it's being relied on by too many.
* TcpFullStack is also removed, pending TcpListen and TcpAccept traits discussed in #69.


Dirbaio:
The UDP traits have the same set of problems (for example, you can only rx/tx on 1 conn at a time). I'd remove them, once they're known-broken there's no point in leaving them.

lulf:
Great, I've removed them.

@MathiasKoch I've also bumped the version to 0.2.0 and added the changelogs.

MathiasKoch:
Released to crates.io :tada:

Dajamante:
Hello Ulf @lulf  and everyone,
Does it mean that ConnectionSocket and ListeningSocket shall be implemented for TCPConnect?
Refering to issue #63!

lulf:
Hi @Dajamante ! I think the discussion in that issue is mainly about the blocking traits. This PR only removed the async traits since I made some bad decisions in the initial design. 

As discussed in #69, maybe TcpListen and TcpAccept traits is the complementary of TcpConnect, but I'm not going to work on that near term due to other priorities, so feel free to have a go :)

PR #76, opened 2022-10-01
Bump no-std-net dependency to 0.6, MSRV to 1.53

chrysn:
The no-std-net version currently used (0.5) failed to parse `SocketAddr` strings `"[fe80::1]%2"`, which works correctly in the 0.6 version.

Incrementing the version is this easy change (nothing inside current embedded-nal appears to depend on the precise version), but it is an API breaking change in the crates here because types are re-exported and their API changed. (For example, `Ipv6Addr::unspecified()` became `Ipv6Addr::UNSPECIFIED`).  Given that these crates have not yet had non-breaking updates, that's probably not a big deal.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
 
         include:
           # Test MSRV
-          - rust: 1.51.0
+          - rust: 1.53.0
             TARGET: x86_64-unknown-linux-gnu
 
           # Test nightly but don't fail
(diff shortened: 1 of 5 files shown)


chrysn:
CI told me that this necessitates an MSRV increase to 1.53, which is now enacted in a second commit.

eldruin (approved):
Looks good to me, thanks!

chrysn:
Thanks for the quick review, that allows me to rebase #73 and saves me workarounds.

PR #77, opened 2022-11-18
async: User async-function-in-trait

chrysn:
This is a WIP attempt to use the recent async-function-in-trait feature. It makes embedded-nal-async regress into nightly-only, but then again the `feature(generic_associated_types)` was not even removed yet, and it's barely possible to actually implement the traits described here without using impl Trait in the associated types' definitions anyway.

## Benefits

* Clearer interfaces
* Better ergonomics for trait implementers
* The provided version of connect() (that falls back to connect_from an unspecified address) is now actually provided

## Downsides

* The lifetimes of the associated types of the stack (`Connected` etc) had to go away due to known issues. But I kind of doubt we'd really have needed them.
  This probably needs further discussion, but before that I'll have to understand again myself why I had them there in the first place.
* async-function-in-trait is an unstable feature (but [it sounds like it would move toward stabilization fast](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html), and it's my impression that it might be there even faster than impl Trait for associated types)

## Status

This is experimentation at the moment, and only done for UDP (thus based on #73 until that is merged; only consider the top few commits for this PR). There is a branch on std-embedded-nal that already followed the switch. [edit: And because GitHub and GitLab don't federate PR mentions like modern issue trackers do: it's on <https://gitlab.com/chrysn/std-embedded-nal/-/tree/async-fn-in-trait>]

Porting DNS and TCP is a step on this PR's path from being a draft PR to a regular one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
+/embedded-nal-async/target
 **/*.rs.bk
 Cargo.lock
--- a/embedded-nal-async/src/lib.rs
+++ b/embedded-nal-async/src/lib.rs
@@ -1,6 +1,7 @@
 //! # embedded-nal-async - An async Network Abstraction Layer for Embedded Systems
 
 #![no_std]
+#![feature(async_fn_in_trait)]
 #![deny(missing_docs)]
 #![deny(unsafe_code)]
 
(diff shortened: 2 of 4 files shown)


chrysn:
Closing this because it's all being included in #73.

The "it's nightly again" and "TCP/DNS is missing" are moot since #78.

I've looked at the Connected (etc) lifetimes, and think that the old lifetimes were just a needless artifact -- so I see no open discussion points worth carrying over to #73.

PR #81, opened 2023-04-22
Add support for ip_in_core

newAM:


embedded-nal-async/Cargo.toml
```diff
@@ -11,6 +11,9 @@ readme = "README.md"
 keywords = ["network"]
 categories = ["embedded", "hardware-support", "no-std", "network-programming", "asynchronous"]
 
+[features]
+ip_in_core = []
```
ryan-summers, line 15: Can't we get rid of no_std_net in this case?

ryan-summers (changes requested):
Minor request on dependencies. However, I'm not sure how meaningful this PR will be until the changes land in stable, since library authors will implement the e-nal, and end users won't have the ability to pick and choose. As such, it seems like optional features are actually bad? Not sure though, what's the e-nal approach?

embedded-nal-async/Cargo.toml
newAM: > Can't we get rid of no_std_net in this case?

Good point! I made it an optional feature.

newAM:
> However, I'm not sure how meaningful this PR will be until the changes land in stable, since library authors will implement the e-nal, and end users won't have the ability to pick and choose.

Crate authors can provide a pass-through to select this feature.  Not a great solution, but at the same time if nobody uses `ip_in_core` then it likely wont get stabilized.

ryan-summers:
Welp, let's try it out and see how it works! :)

ryan-summers (approved):

PR #82, opened 2023-04-22
Update github actions

newAM:
* `actions/checkout@v2` -> `actions/checkout@v3`
  * v2 uses node12 which is deprecated, and soon to be unsupported by github actions
* `actions-rs/toolchain` -> `dtolnay/rust-toolchain`
  * `actions-rs` is unmaintained: https://github.com/actions-rs/toolchain/issues/216

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,13 +27,11 @@ jobs:
             TARGET: x86_64-unknown-linux-gnu
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          target: ${{ matrix.TARGET }}
-          override: true
+          targets: ${{ matrix.TARGET }}
       - uses: actions-rs/cargo@v1
         with:
           command: check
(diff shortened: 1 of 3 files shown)


newAM:
This is untested, it may need a fix or two after the workflows are approved and run :)

newAM:
Huh, I think that clippy failure is just github actions having a bad day.  The http failure 429 is "too many requests", have not seen that one before.

.github/workflows/ci.yml
```diff
@@ -47,13 +45,11 @@ jobs:
         TARGET: [x86_64-unknown-linux-gnu, thumbv6m-none-eabi, thumbv7m-none-eabi]
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
```
ryan-summers, line 49: Can we pin tool chain to a version? I don't like running off master in case it causes random breakage personally

ryan-summers (changes requested):

.github/workflows/ci.yml
newAM: dtolnay implemented it such that the version tags correspond to the toolchain version.  `master` is used when the toolchain version is parametric.

I'm not too concerned about stability since dtolnay has a good record of API stability from serde and proc macros, and I have not had any issues in ~1yr of using it for my own repos.

That all said, I'm happy to find something else if you disagree :)

ryan-summers (approved):
LGTM, we can revisit tags if it's a problem. Better than using deprecated stuff :)

PR #85, opened 2023-05-10
Adding new TCP socket APIs

### LINKED ISSUE #52: TCP socket state/error handling insufficiently specified
The correspondence of the `TcpClientStack` API to TCP socket states is currently not specified clearly enough to be able to write robust TCP client applications against the generic embedded-nal interface. For instance:

* What does `is_connected` mean? That `connect` has been called on the socket? That the socket is ESTABLISHED in TCP terms? That there might still be data to `receive()`? (where the socket could already be in CLOSE-WAIT, but still with data in the receive buffer)
* What does `connect()` returning `Ok()` instead of `WouldBlock` mean? `is_connected() == true` immediately afterwards?
* What happens if the socket is closed gracefully on the remote side? Does `receive()` return `Some(0)` like in the Berkeley sockets API? Or does it return some sort of error? If the latter, how does a client application differentiate between that and unexpected errors?

This is complexity inherent to the problem domain, and needs to be (carefully) exposed in any interface abstracting TCP sockets. Otherwise, it is simply not possible to write robust applications using embedded-nal. For instance, even in the common case where an embedded device just wants to try reconnecting to some server when there are some sort of connection problems, without any care for the exact state of the previous connection, is it okay to try and `connect` the socket again after `is_connected` returns false, or is a new `socket()` necessary?

For an example of a library that has (save for a few typos) done a decent job at this, see https://docs.rs/smoltcp/0.7.0/smoltcp/socket/struct.TcpSocket.html.

ryan-summers:
Fixes #52 by exposing a more robust API for TCP sockets.

I recently [hit an issue with minimq](https://github.com/quartiq/minimq/issues/125) where the `is_connected()` interface was determined to be insufficiently expressive to properly managing the TCP socket state.

### DIFF
--- a/src/stack/tcp.rs
+++ b/src/stack/tcp.rs
@@ -4,7 +4,7 @@ use crate::SocketAddr;
 #[non_exhaustive]
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub enum TcpErrorKind {
-	/// The peer has closed one or both directions and the connection is broken.
+	/// The socket has been closed in the direction in which the failing operation was attempted.
 	PipeClosed,
 
 	/// Some other error has occurred.
--- a/src/stack/tcp.rs
+++ b/src/stack/tcp.rs
@@ -28,20 +28,20 @@ pub trait TcpClientStack {
 	) -> nb::Result<(), Self::Error>;
 
 	/// Determine if a socket is opened.
-    ///
-    /// Returns `Ok(true)` if the TCP socket is actively ingressing and egressing packets. This
-    /// corresponds to any TCP state that is not `CLOSED` or `TIME-WAIT`.
+	///
+	/// Returns `Ok(true)` if the TCP socket is actively ingressing and egressing packets. This
+	/// corresponds to any TCP state that is not `CLOSED` or `TIME-WAIT`.
 	fn is_open(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
 
-    /// Check if the TCP socket can transmit data.
-    ///
-    /// Returns `Ok(true)` if the TCP transmit half is open and connected.
-    fn may_send(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
+	/// Check if the TCP socket can transmit data.
+	///
+	/// Returns `Ok(true)` if the TCP transmit half is open and connected.
+	fn may_send(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
 
-    /// Check if the TCP socket can receive data.
-    ///
-    /// Returns `Ok(true)` if the TCP receive half is open and connected.
-    fn may_receive(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
+	/// Check if the TCP socket can receive data.
+	///
+	/// Returns `Ok(true)` if the TCP receive half is open and connected.
+	fn may_receive(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
 
 	/// Write to the stream.
 	///
--- a/src/stack/tcp.rs
+++ b/src/stack/tcp.rs
@@ -111,8 +111,16 @@ impl<T: TcpClientStack> TcpClientStack for &mut T {
 		T::connect(self, socket, remote)
 	}
 
-	fn is_connected(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error> {
-		T::is_connected(self, socket)
+	fn is_open(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error> {
+		T::is_open(self, socket)
+	}
+
+	fn may_send(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error> {
+		T::may_send(self, socket)
+	}
+
+	fn may_receive(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error> {
+		T::may_receive(self, socket)
 	}
 
 	fn send(
--- a/src/stack/tcp.rs
+++ b/src/stack/tcp.rs
@@ -27,8 +27,21 @@ pub trait TcpClientStack {
 		remote: SocketAddr,
 	) -> nb::Result<(), Self::Error>;
 
-	/// Check if this socket is connected
-	fn is_connected(&mut self, socket: &Self::TcpSocket) -> Result<bool, Self::Error>;
+	/// Determine if a socket is opened.
+    ///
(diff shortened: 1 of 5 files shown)


ryan-summers:
CC @MathiasKoch as I think you're the other user of TCP traits? Also @chrysn

chrysn:
I'm having a hard time following the trouble in the MQTT implementation. What is the decision for which "is this socket in the CONNECTED state" does not suffice?

Do the "may send" / "may receive" / "is open" methods suffice to also cover similar cases, or might we wind up needing to express all the FSM states? Can we get them out of a POSIX socket? Can we even get the ones introduced here on a POSIX system? (I'm generally happy to say "then POSIX sockets are a bad API", but I don't want to do that without due consideration).

MathiasKoch:
I have to say that personally I am currently in the process of migrating towards the async version, and i think that the general API of the async version is MUCH more rust-like, and should be implemented for the blocking version as well if you ask me.

ryan-summers:
I'm back with some findings after trying to use the `embedded-nal-async` approach - thanks for the advice you two :)

> I have to say that personally I am currently in the process of migrating towards the async version, and i think that the general API of the async version is MUCH more rust-like, and should be implemented for the blocking version as well if you ask me.

The main issue that I'm running into is that a struct may not own a `Tcp::Connection` and the `TcpStack` simultaneously.
* This happens because the `Connection` refers to the lifetime of `TcpStack`, which is equivalent to `Self`. Rust does not support self-referential structs

In my use case (MQTT), the MQTT client needs to close and open TCP sockets in response to adverse network events (i.e. link loss, etc.), which implies that it has to own the network stack (in order to establish new connections as required). Because of this, the Connection-based API used by the async version doesn't seem to be the right approach for more complex stack requirements.

---

> I'm having a hard time following the trouble in the MQTT implementation. What is the decision for which "is this socket in the CONNECTED state" does not suffice?

The specific case is differntiating between "attempting to connect", "connected", and "no-longer-connected". All 3 of these cases have different required behaviors, but the `is_connected()` only allows us to detect two states.
1. "attempting to connect" => We should keep waiting and not do anything until we are connected
2. "connected" => We are connected to our peer and can actively rx and tx, send our packets now
3. "no-longer-connected" => We need to close the TCP socket and open a new one

Note that the actions in (1) and (3) are polar opposites. In one case, we actively want to do nothing (i.e. wait for connection). In the other case, we need to close the socket and re-attempt connections.

> Do the "may send" / "may receive" / "is open" methods suffice to also cover similar cases, or might we wind up needing to express all the FSM states?

Honestly, the more I think about it, the more it seems like maybe we should just expose the FSM directly.

 Can we get them out of a POSIX socket? Can we even get the ones introduced here on a POSIX system? (I'm generally happy to say "then POSIX sockets are a bad API", but I don't want to do that without due consideration).

I think POSIX sockets expose this information primarily through returned error codes. If the TX pipe is closed when you try to send, it will error. Similarly with the RX pipe. Maybe this would be a better approach... Let me give it a shot.

chrysn:
> I think POSIX sockets expose this information primarily through returned error codes. If the TX pipe is closed when you try to send, it will error. Similarly with the RX pipe. Maybe this would be a better approach... Let me give it a shot.

That sounds like a EAFP instead of LBYL approach, which I appreciate, so I'm looking forward to seeing what comes out of it.

> own a Tcp::Connection and the TcpStack simultaneously.

IIRC going towards more many-owners-of-shared-stack was a necessity of the async API because multiple components can have pending operations (say, reads) that overlap in time, which would conflict with the necessity to hold an exclusive reference to the stack in the async model. In a blocking model, where one would either read only from one socket or put all one's sockets into some multiplexer (which'd do a read given a single reference to the stack), that's not necessary. I'd hope that many of the nice properties of the async API can be carried over to the sync world, even if the "main argument" is a (&mut socket, &mut stack) instead of a (&mut socket-including-shared-stack).

ryan-summers:
> I'd hope that many of the nice properties of the async API can be carried over to the sync world, even if the "main argument" is a (&mut socket, &mut stack) instead of a (&mut socket-including-shared-stack).

The issue isn't actually mutability, but rather lifetime issues. If you try the following:

```rs
struct MyStruct<T: TcpStack> {
    connection: T::Connection<'?>,
    stack: T,
}
```

What is the lifetime of `connection`? It's supposed to be that of `stack`, but that's not possible to specify because it would be referencing the lifetime of `Self` for `MyStruct`

ryan-summers:
@chrysn Updates are now available on the PR. I opted for a base error type that allows for library users to detect pipe errors that require TCP to be reconnected. It's working well on my local setups and also resolves the original MQTT I was having. Let me know what you think.

If we want to move towards a more socket-directed API, I'd like to keep that off for a new PR. There's a lot of work involved in that

ryan-summers:
@rust-embedded-community/embedded-nal Anyone available to help with review on this one? :)

thejpster:
Seems OK to me, but... Does anyone have a public example showing how they have implemented the new error codes? Can it be done on all the existing stacks, or are you going to break some of them?

ryan-summers:
> Seems OK to me, but... Does anyone have a public example showing how they have implemented the new error codes? 

The above-linked minimq PR is a client side impl (https://github.com/quartiq/minimq/pull/126) and the smoltcp-nal (https://github.com/quartiq/smoltcp-nal/pull/42) above is a stack implementation, both working on hardware with stabilizer.


> Can it be done on all the existing stacks, or are you going to break some of them?

I'd be surprised if a network stack wasn't capable of telling you the socket is closed when trying to read or write to TCP. In any case, a stack doesn't have to implement these error Codes and can mark everything as other if they have to.

thejpster (approved):

src/stack/tcp.rs
```diff
@@ -1,5 +1,22 @@
 use crate::SocketAddr;
 
+/// Represents specific errors encountered during TCP operations.
+#[non_exhaustive]
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub enum TcpErrorKind {
+	/// The peer has closed one or both directions and the connection is broken.
```
chrysn: Could this be more precise, eg. like this?

```suggestion
	/// The socket has been closed in the direction in which the failing operation was attempted.
```

This would
* make no implications on who closed the connection (we currently don't have API to create a half-closed connection from our end, but maybe something else caused the OS to close our stream, not sure if realistic), and
* tell which direction was affected (because one-or-both kind of implies that a write may fail if the inbound half was closed)

chrysn (approved):
LGTM, thanks.

(Not sure if my approval is helpful, but I trust the automation rules know my role here :-) ).

src/stack/tcp.rs
ryan-summers: I like this change for sure! Nice clarification :)

PR #87, opened 2023-06-21
Preparing for a 0.7.0 release

### LINKED ISSUE #84: New release of `embedded-nal`
It seems that there is small discrepancy between the embedded-nal & embedded-nal-async crates which causes some issues for me. embedded-nal current release has dependency on `no-std-net 0.5` but `-async` has on `0.6`.

Even though `no-std-net` hasn't had any release in a while and is not actively maintained, there is a small difference and with the new support for `ip_in_core` it seems fitting to make a new release of `embedded-nal`.

ryan-summers:
Releasing 0.7.0 of the `embedded-nal` to address some TCP soundness issues.

Fixes #84 

I'd also like to take a look at revamping the TCP API to be more like `embassy-net`, but I'm unsure if we should block the release of a new version on potential future work for now.

Thoughts @rust-embedded-community/embedded-nal ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.7.0] - 2023-06-21
 
 * Add blanket impls of all the traits for mutable references.
 - Bump dependency version of `no-std-net` to `v0.6`.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.7.0] - 2023-06-21
```
eldruin, line 10:
```suggestion
## [Unreleased]

## [0.7.0] - 2023-06-21
```
I would keep the unreleased section.

eldruin (commented):
I would not hold the release due to potential future work. We can always make another one.
From the changelog, it seems there is enough stuff unreleased.

For this to be merged, though, the dependency in the async crate needs to be updated, which is now failing on CI.

ryan-summers:
Deps for e-n-async should now be updated. Thanks for pointing that out, didn't realize the async version depended on the sync version

thejpster:
Little and often. This is the way.

eldruin (approved):
Thank you!

PR #93, opened 2023-11-08
async: Do not use heapless on public API.

Dirbaio:
It's better to not use heapless on the public API.

- It can get major bumps from time to time (for example v0.8 was just relased), and upgrading it is a breaking change of embedded-nal-async, which we should try to avoid.
- It allows users to use any version of heapless, or something else entirely.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [ staging, trying, master ]
+    branches: [staging, trying, master]
   pull_request:
 
 name: Continuous integration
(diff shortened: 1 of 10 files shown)


lulf (approved):
Fully agreed! This bothered me with the existing blocking api, should that be changed too?

ryan-summers:
Would you be willing to copy these over to the non-async version as well? I agree with the sentiment here, and we have the same problem over there.

MathiasKoch (approved):

Dirbaio:
@ryan-summers  done

also note I've separated the MSRV bump to #94, review that first.

ryan-summers (approved):
Thanks for fixing this!

PR #99, opened 2023-11-28
Release nal v0.8.1, nal-async v0.7.1

jannic:


jannic:
Didn't notice that the feature change only affects embedded-nal-async

PR #100, opened 2023-12-05
Change blocking Dns::get_host_by_address back to mutable self

MathiasKoch:
This aligns the signature of the blocking Dns trait with get_host_by_name, by re-adding the `&mut self` and `nb::Error` result type.

It used to be this way, but it seems like https://github.com/rust-embedded-community/embedded-nal/pull/93# was a bit too fast on copying over the async api.

Ping @Dirbaio


ryan-summers (approved):
Indeed, thanks for catching this one :)

Dirbaio (approved):
sounds good to me. sorry for the screw-up :see_no_evil:

PR #102, opened 2023-12-29
Use stabilized ip_in_core feature

newAM:
`ip_in_core` is stabilized in rust 1.77, I think it makes sense to use `core::net` instead of `no-std-net` when 1.77 reaches stable on 2024-03-21.

This is a draft PR to start the discussion.  I will update CI to run on stable when 1.77 releases.


MathiasKoch (approved):

ryan-summers (approved):
Woohoo! Finally! Thanks for the changes :)

Issue #103, opened 2024-02-05
e-nal-async: Missing lifetimes in `UdpStack`


ivmarkov:
`Connected`, `UniquelyBound` and `MultiplyBound` [associated types](https://github.com/rust-embedded-community/embedded-nal/blob/master/embedded-nal-async/src/stack/udp.rs#L126) are currently *not* lifetimed.

I need them to be, so that I can keep a non-mutable reference to the "Udp Stack" instance in their impls.
While I can `Arc` the "Udp Stack" and then push the arc into the UDP socket impl, this just doesn't feel right.

Perhaps it is no coincidence that embassy-net's UDP socket stack is not implementing the e-nal-async traits? I don't think it would be possible anyway, without the above lifetimes.

Background: I stumbled on this by accident, while trying to implement a toy websockets gateway that tries to implement `embedded-nal-async` traits by multiplexing all traffic (UDP and TCP) over a single websocket to an actual STD based `embedded-nal-async` impl.

Anyway, I'll also work on a PR that implements a lifetimed version of the UDP associated types on top of `embassy-net`.

ivmarkov:
Supporting the [`bind_multiple`](https://github.com/rust-embedded-community/embedded-nal/blob/master/embedded-nal-async/src/stack/udp.rs#L198) contract on top of `smoltcp` seems currently impossible, or in the best case would require terrible hacks, as `smoltcp` does seem to assign a source (local) address *very late* - [only when dispatching the UDP packet](https://github.com/smoltcp-rs/smoltcp/blob/main/src/socket/udp.rs#L521). The UDP packets in the ring buffer queue however do not have any notion of "source address" before that. 

So it seems sending with an arbitrary local address would require us to 
a) wait for the UDP socket ring buffer to become empty (so that we are sure all other pending packets are sent with the "previous" local addr)
b) unbind the UDP socket
c) rebind the socket with the local address of the packet we want to send (NOTE: this way **however** we would be missing packets that we want to receive, as the `receive_from` part will now work with the local address that we want to use for *sending* the packet
d) unbind the socket once the packet was sent
e) rebind the socket again with the local address as specified in `bind_multiple`

[...]

Issue #92, opened 2023-10-22
SocketAddr associated types


chrysn:
Implementing embeded-hal(-async), it seems I'm always converting SocketAddr back and forth with the platform's own socket addr type.

Would it make sense to add an associated SocketAddr type (probably constrained to `Into<SocketAddr> + TryFrom<SocketAddr, Error=Unavailable> + PartialEq<Self::SocketAddr> + Hash`)? In all places where we previously pass SocketAddr, we'd now pass Self::SocketAddr. Wherever users store addresses, they'd use their stack's SocketAddr type, and only convert through a SocketAddr when entering or leaving the stack (or even not at all if we require Display and FromStr).

Some reasons why a stack would not like working with SocketAddr all the time:
* It is built on something where Rust's SocketAddr is just not the native format (eg. any non-Rust network stacks), so it needs to copy fields around all the time. (This is the case with RIOT OS)
* It is an IPv6 only stack that doesn't constantly want to deal with erring out on IPv4 addresses.
* It is an IPv4/6 dual stack that internally represents IPv4 addresses as IPv4-mapped addresses.
* It has only a single network interface and thus doesn't keep a zone identifier around. (In RIOT, there are some optimizations for what is called "highlander mode" when there can be only one).

This is mainly a concern for users of unconnected UDP sockets, where recvfrom-style use is predominant, and addresses are handled in every single packet. For TCP or connected UDP sockets, it's probably a secondary concern.

ryan-summers:
I am very okay with a generic socketaddr type myself, as I've had these exact same issues during implementation. I don't see any specific benefit to using `no-std-net`, since we still have versioning issues, and there's not very convenient methods to convert between common types.

ryan-summers:
However, how would a library that is agnostic to the underlying stack know how to construct and use these addresses? I.e. https://github.com/quartiq/minimq/blob/master/src/network_manager.rs, where we don't know what stack we're running on. Would this be the `TryFrom` impl bound?

chrysn:
[...]

PR #106, opened 2024-02-07
UDP stack compatible with embassy-net; socket splitting

ivmarkov:
Addresses #103 

As per the subject, this PR contains two changes (I can split into two separate PRs as well, but given that both would be backwards incompatible, and are touching the same traits, perhaps it is better to discuss those in one go):

* The PR introduces lifetimes in the `UdpStack::Connected`, `UdpStack::UniquelyBound` and `UdpStack::MultiplyBound` associated types, making those GATs
  * The change is similar and following the lead as to how the the `TcpConnect::Connection` type [is modeled already](https://github.com/rust-embedded-community/embedded-nal/blob/master/embedded-nal-async/src/stack/tcp.rs#L12)
  * It is necessary, so that the e-nal-async `UdpStack` trait is implementable in `embassy-net` and possibly other non-allocating network stacks using an [identical approach](https://github.com/ivmarkov/embassy/blob/master/embassy-net/src/udp.rs#L239) to the [already implemented](https://github.com/embassy-rs/embassy/blob/main/embassy-net/src/tcp.rs#L610) `TcpConnect` factory trait
* The second change in this PR is that it puts an extra requirement on the `UdpSocket` factory trait. Namely, that the sockets it returns should be *splittable* into two separate halves - "receive" and "send" which can be used concurrently in app protocols that do require full-duplex UDP datagram sending and receiving. This is done by:
  * Splitting the `ConnectedUdp` and `UnconnectedUdp` existing traits into separate "send" and "receive" traits: `ConnectedUdpReceive`, `ConnectedUdpSend`, `UnconnectedUdpReceive` and `UnconnectedUdpSend` - similarly as to how the `TcpConnect::Connection` associated type is constrained by two separate traits for reading and writing from/to TCP sockets - namely - `embedded_io_async::Read` and `embedded_io_async::Write`
  * Introducing two new traits: `ConnectedUdpSplit` and `UnconnectedUdpSplit` - which model the notion of a splittable connected/unconnected UDP socket by using `Send/Receive` lifetimed associated types.

Background on these changes, and justification for introducing UDP socket splitting can be found in #103 

Open topics / next steps:
* As per #103 I think we should come to an agreement that socket splitting is important and necessary.

* Regardless of the outcome of the above, I think the separation of `ConnectedUdp` and `UnconnectedUdp` into "receive" and "send" parts **stands on its own**. Reason being - if a concrete network stack does allow UDP socket splitting, and even if we decide (regrettably) not to require this in the `UdpSocket` factory, app code that **does** need split UDP sockets would still benefit from the existence of separate "send" and "receive" traits, for the very same reasons we have separate `Read` and `Write` traits for TCP / streaming cases - i.e. the app code can take an "already splitted" in a platform-specific way UDP socket in the form of - e.g. `async fn foo(send: impl UnconnectedUdpSend, recv: impl UnconnectedUdpReceive)` and then proceed with polling these separately

* (**UPDATED**) The existing traits' names are mouthful. How about the following simplification:
  * Rename:
    * `UnconnectedUdpReceive` -> `UdpReceive`
    * `UnconnectedUdpSend` -> `UdpSend`
    * `UnconnectedUdpSplit` -> `UdpSplit`
    * Retire ALL of the `Connected*` traits and use the `Unconnected*` ones for the connected case as well. We are anyway already a bit in this situation, by re-using the `Unconnected*` traits for "single-bound" sockets, even though the API signatures of the unconnected traits are not a perfect fit there, as they take / return the `local` socket addr, which serves no useful purpose when the socket is already single-bound to a fixed local IP address.

### DIFF
--- a/embedded-nal-async/src/lib.rs
+++ b/embedded-nal-async/src/lib.rs
@@ -19,4 +19,7 @@ pub use no_std_net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, Socke
 pub use dns::Dns;
 pub use embedded_nal::AddrType;
 pub use stack::TcpConnect;
-pub use stack::{ConnectedUdp, UdpStack, UnconnectedUdp};
+pub use stack::{
+	ConnectedUdpReceive, ConnectedUdpSend, ConnectedUdpSplit, UdpStack, UnconnectedUdpReceive,
+	UnconnectedUdpSend, UnconnectedUdpSplit,
+};
--- a/embedded-nal-async/src/stack/mod.rs
+++ b/embedded-nal-async/src/stack/mod.rs
@@ -2,4 +2,7 @@ mod tcp;
 mod udp;
 
 pub use tcp::TcpConnect;
-pub use udp::{ConnectedUdp, UdpStack, UnconnectedUdp};
+pub use udp::{
+	ConnectedUdpReceive, ConnectedUdpSend, ConnectedUdpSplit, UdpStack, UnconnectedUdpReceive,
+	UnconnectedUdpSend, UnconnectedUdpSplit,
+};
--- a/embedded-nal-async/src/stack/udp.rs
+++ b/embedded-nal-async/src/stack/udp.rs
@@ -14,9 +14,11 @@
 //!   Implementing `UniquelyBound` and `MultiplyBound` with the same type is expected to be a
 //!   common choice.
 
+use embedded_io_async::ErrorType;
+
 use crate::SocketAddr;
 
-/// This trait is implemented by UDP sockets.
+/// This trait is implemented by UDP sockets and models their datagram receiving functionality.
 ///
 /// The socket it represents is both bound (has a local IP address, port and interface) and
 /// connected (has a remote IP address and port).
@@ -26,13 +28,7 @@ use crate::SocketAddr;
 /// of establishing a connection (which is absent in UDP). While there is typically no POSIX
 /// `bind()` call in the creation of such sockets, these are implicitly bound to a suitable local
 /// address at connect time.
-pub trait ConnectedUdp {
-	/// Error type returned by send and receive operations.
-	type Error: embedded_io_async::Error;
-
-	/// Send the provided data to the connected peer
-	async fn send(&mut self, data: &[u8]) -> Result<(), Self::Error>;
-
+pub trait ConnectedUdpReceive: ErrorType {
 	/// Receive a datagram into the provided buffer.
 	///
 	/// If the received datagram exceeds the buffer's length, it is received regardless, and the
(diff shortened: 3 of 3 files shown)


ivmarkov:
@chrysn @MathiasKoch Would appreciate comments / feedback here. :) As in, am I on the right track? any concerns / suggestions?

chrysn:
I think the lifetime change is on the right track.

The splitting worries me -- while so far RIOT sends instantly, once they allow network interface backpressure (there were bugs due to the lack of it), splitting becomes messy at best.

Unified types IMO would require per-type send and receive AT from <https://github.com/rust-embedded-community/embedded-nal/issues/92> -- so lets do that :-)

ivmarkov:
> The splitting worries me -- while so far RIOT sends instantly, once they allow network interface backpressure (there were bugs due to the lack of it), splitting becomes messy at best. 

The problem with *not* enhancing `UdpStack` to return `UdpSplit` types is that folks wouldn't be able to use this factory for any non-request-response app protocol that lives on top of UDP, which (severely, IMO) limits its applicability.

Worse, if we don't at least have separate `UdpSend` and `UdpReceive` traits (similarly to `embedded-io-async` `Read` and `Write`) and continue to live with a single send+receive `*Udp` trait, folks in my situation wouldn't be able to use even the UDP socket trait - *without* the `UdpStack` factory - in their apps. As in - I don't see *what* aspect of the UDP functionality of `embedded-nal-async` I would be able to actually use in my mDNS and Matter use cases then?

Why is implementing splittable sockets on top of RIOT so complicated? Even if RIOT ends up having a *single* callback for both "you can try sending now" and "got incoming packet" and thus with a single waker registration slot, [you could still split the socket and it should work fine for intra-task concurrency (i.e. where you use the send and receive halves within a single task)](https://github.com/rust-embedded-community/embedded-nal/issues/103#issuecomment-1931778363) - without any increased CPU consumption whatsoever.
 
> Unified types IMO would require per-type send and receive AT from <#92> -- so lets do that :-)

If I understand you correctly, you are suggesting - additionally to the changes suggested here and **including** the change where we end up with a single `Udp(Send)/(Receive)` (pair) of traits - to also enhance these traits with its own `SocketAddr` associated type?

If yes, then I can do that, but even before that, if you could comment on how do you feel of having a single `Udp` (or as per above and better yet - a single `UdpReceive` / `UdpSend` pair) for **all** of connected / multiply-bound / single-bound use cases even though `local`, `remote` or both socket addrs might be rendundant in 2 out of the 3 use cases?

chrysn:
> The problem with *not* enhancing `UdpStack` to return `UdpSplit` types
> is that ...

I'm not fundamentally opposed to this split -- I'll merely need to look
at good examples of how this would be implemented in stacks such as
RIOT, and I'll want to understand better how it applies to protocols
such as mDNS which have request-response components but also independent
transmissions. (Like, will applications require that the sender is Clone
as well?).

Therefore, I'd propose to treat the concerns separately, as the lifetime
change should be something that can be approved fast (I don't have the
position to do it but can review).

> If I understand you correctly, you are suggesting - additionally to
> the changes suggested here and **including** the change where we end
> up with a single `Udp(Send)/(Receive)` (pair) of traits - to also
> enhance these traits with its own `SocketAddr` associated type?

Rougly so, yes. Don't have a full plan yet, but the addresses would be
TryFrom and Into core::net::SocketAddress (stable in 1.77, conveniently
resolving an issue that so far blocked #92). They could be lifetime
generic, and thus allow addresses just be newtypes around &Socket for
connected sockets. Given that the API would look like UnconnectedSocket
interfaces, there might even be a single AT for the (local, remote)
tuple.

> If yes, then I can do that, but even before that, if you could comment
> on how do you feel of having a single `Udp` (or as per above and
> better yet - a single `UdpReceive` / `UdpSend` pair) for **all** of
> connected / multiply-bound / single-bound use cases even though
> `local`, `remote` or both socket addrs might be rendundant in 2 out of
> the 3 use cases?

Yes, let's do that. Once we have ATs per socket, the reason for having
Connected (and the half other type) goes away, as that was about not
always passing around large addresses -- but for a connected type we
wouldn't need to if its AT is as small as it can be.

chrysn:
The Send/Receive split does sound like a good direction to go in, I've recently been shown another example of where it is convenient (although might really be the same example; it's rs-matter's mDNS server).

A colleague is working on imlementing split sockets on top of the current embedded-nal sockets. I don't think it'll be super efficient, as there may be a few synchronization primitives involved, but if such an implementation can be made, if any OS does *not* allow selecting for read and write separately (or placing separate callbacks for read and write ready), it can still use such a solution.

ivmarkov:
> A colleague is working on imlementing split sockets on top of the current embedded-nal sockets. I don't think it'll be super efficient, as there may be a few synchronization primitives involved, but if such an implementation can be made, if any OS does _not_ allow selecting for read and write separately (or placing separate callbacks for read and write ready), it can still use such a solution.

Something like [this](https://github.com/ivmarkov/edge-net/blob/master/edge-mdns/src/io/split.rs)?

Unfortunately only possible for UDP. I don't think that approach would work for streaming protocols like TCP. Not with the current `async fn` specification of the UDP send/receive and `embedded_io_async` read/write methods that is.

Also, the price is less convenience and one extra buffer.

A generic split implementation that does not need extra buffers and works for TCP as well is only possible if the `Read`/`Write` traits are defined in terms of the old `poll` based API, like the STD async read/write traits in `futures`, `tokio` and so on. IMO. But if you manage to pull it through somehow - I'm all ears. :)

ivmarkov:
> The Send/Receive split does sound like a good direction to go in, I've recently been shown another example of where it is convenient (although might really be the same example; it's rs-matter's mDNS server).

The mDNS responder in `rs-matter` and `edge-net` are the same, yes, modulo some insignificant differences. Both were created by me and @bjoernQ

chrysn:
> Something like [this](https://github.com/ivmarkov/edge-net/blob/master/edge-mdns/src/io/split.rs)?

The splitter would not even need to store a packet. It'd suffice if there were a channel by which the recver and the sender can communicate and where they can share the underlying exclusive reference. The sender would attempt to take the shared reference, and on failure signals the recver to please relinquish control. The receiver would do recv racing against reception through the shared channel, and if the channel wins, cancell the receive, return the exclusive reference so that the sender can take it, and retry after the sender is done. Of course that only works for cancellable futures, but AIU it's the understanding that that is the case here. (May want to work on that in parallel).

Let's see how the implementation works out. As this is at least all more complex than the lifetime introduction, would you mind splitting that out, so that it can progress further?

> [same] mDNS responder

I'm still having a hard time wrapping my head around why that particular application works better with the split. Granted, my [CoAP implementation](https://crates.io/crates/embedded-nal-coap) is more request-response style, but it implements both a server and a client at the same time, and while it is not implemented here, it could easily be extended to allow requests with no processed response (it's in the protocol, just not the implementation). Gotta have a closer look at the mDNS responder…

PR #109, opened 2025-10-01
Update embedded-io, release.

Dirbaio:
- **Update embedded-io to 0.7**
- **Release embedded-nal-async v0.9.0**


chrysn (approved):
Remarks from during review (no action needed, merely for documentation):
* This only updates -async because only its TCP traits build on embedded-io; the sync version uses completely different (more socket-style) idioms.
* As we're depending on the trait, supporting both versions (like if this crate just took an impl by argument) is not an option.

Build-tested (also runtime tested but its examples don't exactly cover the changed paths) with std-embedded-nal-async (to be updated next).

